PR #25, opened 2024-11-25
GH-4: [CI] Enable Dependabot

### LINKED ISSUE #4: Enable Dependabot
We only need this part: https://github.com/apache/arrow/blob/a6fe59592c643d591ee719041dd2b5d8b9ada7d6/.github/dependabot.yml#L20-L33

lidavidm:
Fixes #4.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,33 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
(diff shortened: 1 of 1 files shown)


raulcd (approved):
LGTM

.github/dependabot.yml
```diff
@@ -0,0 +1,33 @@
...
+      interval: "weekly"
+    commit-message:
+      prefix: "MINOR: [CI] "
+    open-pull-requests-limit: 10
+  - package-ecosystem: "maven"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    commit-message:
+      prefix: "MINOR: [Java] "
```
kou, line 32: We may not need `[Java]` tag because apache/arrow-java has only the Java implementation.

lidavidm: Ah, I'll fix it, good point

PR #39, opened 2024-11-26
GH-5: [Docs] Add LICENSE and NOTICE files

### LINKED ISSUE #5: Add LICENSE.txt and NOTICE
See also: https://www.apache.org/legal/src-headers.html

raulcd:
Closes #5


raulcd:
I've trimmed down and removed all dependencies that seem to refer to C++ / Python / and other implementations and left the ones that are related to Java (or seem to be related to Java).

NOTICE.txt
```diff
@@ -0,0 +1,13 @@
...
+This product includes software developed at
+The Apache Software Foundation (http://www.apache.org/).
+
+This product includes software from Dremio (Apache 2.0)
+  * Copyright (C) 2017-2018 Dremio Corporation
+  * https://github.com/dremio/dremio-oss
+
+This product includes software from Google Guava (Apache 2.0)
+  * Copyright (C) 2007 The Guava Authors
+  * https://github.com/google/guava
```
kou, line 13: Can we move this to `LICENSE.txt`?

https://infra.apache.org/licensing-howto.html#alv2-dep

> Bundling an Apache 2-0-licensed dependency
>
> Assuming that the bundled dependency itself contains no bundled sub-components under other licenses, so the ALv2 applies uniformly to all files, there is no need to modify LICENSE. However, for completeness it is useful to list the products and their versions, as is done for products under other licenses.
>
> If the dependency supplies a NOTICE file, its contents must be analyzed and the relevant portions bubbled up into the top-level NOTICE file.

raulcd: I am ok moving it to `LICENSE.txt` but to be honest I am unsure I understand what the `NOTICE.txt` refers then when it says, here https://infra.apache.org/licensing-howto.html#mod-notice:
```
Copyright notifications which have been relocated, rather than removed, from source files must be preserved in NOTICE. However, elements such as the copyright notifications embedded within BSD and MIT licenses do not need to be duplicated in NOTICE. You can leave those notices in their original locations.
```
or here: https://www.apache.org/legal/src-headers.html#notice
```
Use the remainder of the NOTICE file to include [required third-party notices][1]. The NOTICE file may also include copyright notices moved from source files submitted to the ASF.
```

LICENSE.txt
```diff
@@ -0,0 +1,223 @@
...
+
+These file are derived from code from Netty, which is made available under the
+Apache License 2.0.
+
+--------------------------------------------------------------------------------
+
+This product includes software from Dremio, which is made available under the
+Apache License 2.0.
+  * Copyright (C) 2017-2018 Dremio Corporation
+  * https://github.com/dremio/dremio-oss
```
kou, line 216: It seems that we can remove this because we removed "Copyright (C) 2017-2018 Dremio Corporation" by https://github.com/apache/arrow/pull/5602 .
Anyway, I should have asked why we can remove it when I merge...

raulcd: Thanks @kou ! I've removed it

raulcd:
I've rebased main but it seems pre-commit is failing there. I've opened:
- https://github.com/apache/arrow-java/issues/433

LICENSE.txt
```diff
@@ -0,0 +1,218 @@
...
+
+--------------------------------------------------------------------------------
+vector/src/main/java/org/apache/arrow/vector/util/IntObjectHashMap.java
+vector/src/main/java/org/apache/arrow/vector/util/IntObjectMap.java
+
+These file are derived from code from Netty, which is made available under the
+Apache License 2.0.
+
+--------------------------------------------------------------------------------
+
```
kou, line 212:
```suggestion
```

kou (approved):
+1

PR #339, opened 2024-11-27
GH-6: Add Linux test CI

### LINKED ISSUE #6: Add test CI: Docker based
This part: https://github.com/apache/arrow/blob/a6fe59592c643d591ee719041dd2b5d8b9ada7d6/.github/workflows/java.yml#L59-L111

We don't need to use Archery like apache/arrow-go.

lidavidm:
Fixes #6.

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,52 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# All of the following environment variables are required to set default values
+# for the parameters in docker-compose.yml.
+
+# empty prefix means that the docker-compose configuration will use named
+# volumes which potentially improves the performance on docker for macos and
+# docker for windows, it also prevents the contamination of the source
+# directory
+# a non-empty prefix means that directories from the host are bind-mounted
+# into the container, it should be set to ".docker/" on github actions to keep
+# the cache plugin functional
+DOCKER_VOLUME_PREFIX=
+
+# turn on inline build cache, this is a docker buildx feature documented
+# at https://github.com/docker/buildx#--cache-tonametypetypekeyvalue
+BUILDKIT_INLINE_CACHE=1
+COMPOSE_DOCKER_CLI_BUILD=1
+DOCKER_BUILDKIT=1
+
+# different architecture notations
+ARCH=amd64
+ARCH_ALIAS=x86_64
+ARCH_SHORT=amd64
+
+# Default repository to pull and push images from
+REPO=apache/arrow-dev
+
+# The setup attempts to generate coredumps by default, in order to disable the
+# coredump generation set it to 0
+ULIMIT_CORE=-1
+
+# Default versions for platforms
+
+# Default versions for various dependencies
+JDK=11
+MAVEN=3.9.6
--- a/.env
+++ b/.env
@@ -39,13 +39,11 @@ ARCH_ALIAS=x86_64
 ARCH_SHORT=amd64
 
 # Default repository to pull and push images from
-REPO=apache/arrow-dev
+REPO=ghcr.io/apache/arrow-java-dev
 
 # The setup attempts to generate coredumps by default, in order to disable the
 # coredump generation set it to 0
 ULIMIT_CORE=-1
-
-# Default versions for platforms
 
 # Default versions for various dependencies
 JDK=11
--- a/.github/workflows/java.yml
+++ b/.github/workflows/java.yml
@@ -53,12 +53,12 @@ jobs:
       MAVEN: ${{ matrix.maven }}
     steps:
       - name: Checkout Arrow
-        uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # v4.0.0
+        uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4.2.1
         with:
           fetch-depth: 0
           submodules: recursive
       - name: Cache Docker Volumes
-        uses: actions/cache@13aacd865c20de90d75de3b17ebe84f7a17d57d2 # v4.0.0
+        uses: actions/cache@6849a6489940f00c2f30c0fb92c6274307ccb58a # v4.1.2
         with:
           path: .docker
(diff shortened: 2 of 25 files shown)


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "testing"]
+	path = testing
+	url = git@github.com:apache/arrow-testing
```
kou, line 3:
```suggestion
	url = https://github.com/apache/arrow-testing.git
```

.github/workflows/java.yml
kou, line 1: How about `test.yml` instead of `java.yml`?

.env
```diff
@@ -0,0 +1,52 @@
...
+ARCH_SHORT=amd64
+
+# Default repository to pull and push images from
+REPO=apache/arrow-dev
+
+# The setup attempts to generate coredumps by default, in order to disable the
+# coredump generation set it to 0
+ULIMIT_CORE=-1
+
+# Default versions for platforms
```
kou, line 48: Can we remove this?

.env
```diff
@@ -0,0 +1,52 @@
...
+COMPOSE_DOCKER_CLI_BUILD=1
+DOCKER_BUILDKIT=1
+
+# different architecture notations
+ARCH=amd64
+ARCH_ALIAS=x86_64
+ARCH_SHORT=amd64
+
+# Default repository to pull and push images from
+REPO=apache/arrow-dev
```
kou, line 42: How about using `ghcr.io/apache/arrow-java-dev`?
If we use Docker Hub, we need to set Docker Hub secrets.

.github/workflows/java.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Checkout Arrow
+        uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # v4.0.0
+        with:
+          fetch-depth: 0
+          submodules: recursive
+      - name: Cache Docker Volumes
+        uses: actions/cache@13aacd865c20de90d75de3b17ebe84f7a17d57d2 # v4.0.0
+        with:
+          path: .docker
+          key: maven-${{ hashFiles('java/**') }}
```
kou, line 64: Could you update the `hashFiles()` argument?

.github/workflows/java.yml
```diff
@@ -0,0 +1,73 @@
...
+      matrix:
+        jdk: [11, 17, 21, 22]
+        maven: [3.9.6]
+        image: [java]
+    env:
+      JDK: ${{ matrix.jdk }}
+      MAVEN: ${{ matrix.maven }}
+    steps:
+      - name: Checkout Arrow
+        uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # v4.0.0
```
kou, line 56:
```suggestion
        uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4.2.1
```

.github/workflows/java.yml
```diff
@@ -0,0 +1,73 @@
...
+        with:
+          fetch-depth: 0
+          submodules: recursive
+      - name: Cache Docker Volumes
+        uses: actions/cache@13aacd865c20de90d75de3b17ebe84f7a17d57d2 # v4.0.0
```
kou, line 61:
```suggestion
        uses: actions/cache@6849a6489940f00c2f30c0fb92c6274307ccb58a # v4.1.2
```

.github/workflows/java.yml
```diff
@@ -0,0 +1,73 @@
...
+        with:
+          path: .docker
+          key: maven-${{ hashFiles('java/**') }}
+          restore-keys: maven-
+      - name: Execute Docker Build
+        env:
+          DEVELOCITY_ACCESS_KEY: ${{ secrets.GE_ACCESS_TOKEN }}
```
kou, line 68: Do we need to set this secret?

arrow-format/README.rst
```diff
@@ -0,0 +1,25 @@
...
+.. under the License.
+
+Arrow Protocol Files
+====================
+
+This folder contains binary protocol definitions for the Arrow columnar format
+and other parts of the project, like the Flight RPC framework.
+
+For documentation about the Arrow format, see the `docs/source/format`
+directory.
```
kou, line 25: How about mentioning that this is a copy of https://github.com/apache/arrow/tree/main/format ?
We may want to mention the revision of the copy.

.github/workflows/java.yml
lidavidm: Yes, we'll need to ask Infra to set it up again - this implements build caching (I'll have to figure out the details). It's not strictly required, though

.github/workflows/test.yml
```diff
@@ -0,0 +1,74 @@
...
+          restore-keys: maven-${{ matrix.jdk }}-${{ matrix.maven }}-
+      - name: Execute Docker Build
+        env:
+          # Enables build caching, but not strictly required
+          DEVELOCITY_ACCESS_KEY: ${{ secrets.GE_ACCESS_TOKEN }}
+        run: |
+          docker compose run \
+            -e CI=true \
+            -e "DEVELOCITY_ACCESS_KEY=$DEVELOCITY_ACCESS_KEY" \
+            ${{ matrix.image }}
```
kou, line 74: We need `docker compose pull`/`docker compose push` to cache built images but let's work on it as a separated task.

kou (approved):
+1

.github/workflows/test.yml
```diff
@@ -0,0 +1,74 @@
...
+      - name: Checkout Arrow
+        uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4.2.1
+        with:
+          fetch-depth: 0
+          submodules: recursive
+      - name: Cache Docker Volumes
+        uses: actions/cache@6849a6489940f00c2f30c0fb92c6274307ccb58a # v4.1.2
+        with:
+          path: .docker
+          key: maven-${{ matrix.jdk }}-${{ matrix.maven }}-${{ hashFiles('**/docker-compose.yml') }}
```
kou, line 64: We may want to update cache by `.java` changes. Let's work on it as a separated task.

lidavidm: Hmm, since the docker image is just jdk/maven, shouldn't it not depend on the actual Java sources?

kou: I think that this is used to cache `mvn ...` results. So the cache will depend on `*.java` too.

.github/workflows/test.yml
kou: Sorry. We don't need this. Because we use existing Docker images...

We can remove `REPO` from `.env` too.

.github/workflows/test.yml
lidavidm: Hmm, the build script starts by wiping the build directory, though. (And we have separate build caching support.)

kou: Oh, I'm not familiar with Maven cache...

This `.docker` path caches `~/.m2/` directory. Do we need to cache `~/.m2/`? If we don't need to cache `~/.m2/`, we can remove this cache.

lidavidm: Ah, this is confusing then. If it caches `.m2` then it'll cache downloaded dependencies, which is useful, and that depends on `pom.xml` (not `.java` though)

lidavidm: I'll update this to hash `pom.xml` in the new PR

PR #430, opened 2024-11-28
GH-10: Add basic pre-commit-config with Apache RAT

### LINKED ISSUE #10: pre-commit: ShellCheck
See also: https://github.com/apache/arrow-go/blob/e112ad0de5ca66133256f2db71bda4dd2d1731f5/.pre-commit-config.yaml#L46-L49

lidavidm:
Fixes #10.


lidavidm:
Ah, I should add the CI check too

kou (approved):
+1

lidavidm:
Weird, it seems `--all-files` does not pass in other PRs, let me see...

lidavidm:
I think the merge didn't account for pre-commit, should hopefully be fixed in #432

.pre-commit-config.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+  - repo: local
+    hooks:
+      - id: rat
+        name: Release Audit Tool
+        language: system
+        entry: |
+          bash -c " \
+            git archive HEAD \
+              --prefix=apache-arrow-go/ \
```
kiszk, line 37: I am not familiar with this tool. But is it better to use `java` instead of `go`?

kou: Good catch!
We should use "java" here.

lidavidm: Ah sorry, I copy-pasted and didn't check :sweat_smile:

PR #449, opened 2024-12-05
GH-13: Set up JNI build (dataset, etc.)

### LINKED ISSUE #13: Add test CI: JNI
This part: https://github.com/apache/arrow/blob/main/dev/tasks/java-jars/github.yml

lidavidm:
Fixes #13.


lidavidm:
The docker image takes so long to build that we really need to cache it.

lidavidm:
It's going to be a while before I can get back to this.

lidavidm:
Hmm, this problem doesn't happen on a local build if I change to a debug build...

lidavidm:
Ah, interesting, LD_PRELOADING ASan (even if the binaries are built without it) "fixes" the issue. That's rather unfortunate, since ASan was how I was trying to figure out the corruption issue...

lidavidm:
Valgrind also seems to "fix" the issue :/

lidavidm:
I may disable the ORC test for now to get CI passing...we'll have to revisit a lot of the JNI work.

wgtmac:
What is the error message for ORC test?

lidavidm:
It is crashing in CI. When I debug locally it appears that a malloc assertion fails. However it no longer reproduces for me after fiddling around.

lidavidm:
I tried to use ASan and Valgrind (separately) to identify possible memory corruption but it turns out under these tools, the crash no longer reproduces. Also after using the tools, even with them turned off now, I can't reproduce the crash locally anymore.

wgtmac:
I suspect ORC crash is due to missing the timezone database: https://github.com/apache/arrow/issues/36026.

lidavidm:
It crashes in malloc, though. I do have the core dump:

```
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at ./nptl/pthread_kill.c:44
#1  0x00007f6268d37f1f in __pthread_kill_internal (signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
#2  0x00007f6268ce8fb2 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007f6268cd3472 in __GI_abort () at ./stdlib/abort.c:79
#4  0x00007f6268d2c430 in __libc_message (action=action@entry=do_abort, fmt=fmt@entry=0x7f6268e48b80 "Fatal glibc error: malloc assertion failure in %s: %s\n")
    at ../sysdeps/posix/libc_fatal.c:155
#5  0x00007f6268d442bc in __malloc_assert (function=0x7f6268e49ba0 <__PRETTY_FUNCTION__.8> "sysmalloc", line=2611, file=<synthetic pointer>, 
    assertion=0x7f6268e49330 "(old_top == initial_top (av) && old_size == 0) || ((unsigned long) (old_size) >= MINSIZE && prev_inuse (old_top) && ((unsigned long) old_end & (pagesize - 1)) == 0)") at ./malloc/malloc.c:299
#6  sysmalloc (nb=nb@entry=2624, av=av@entry=0x7f6260000030) at ./malloc/malloc.c:2611
#7  0x00007f6268d4508e in _int_malloc (av=av@entry=0x7f6260000030, bytes=bytes@entry=2614) at ./malloc/malloc.c:4403
#8  0x00007f6268d45989 in __GI___libc_malloc (bytes=2614) at ./malloc/malloc.c:3323
#9  0x00007f61321523bf in orc::DataBuffer<char>::reserve(unsigned long) () from /tmp/target7371369939985176724arrow_orc_jni
#10 0x00007f6132152426 in orc::DataBuffer<char>::DataBuffer(orc::MemoryPool&, unsigned long) () from /tmp/target7371369939985176724arrow_orc_jni
#11 0x00007f613214c9dc in orc::createReader(std::unique_ptr<orc::InputStream, std::default_delete<orc::InputStream> >, orc::ReaderOptions const&) ()
   from /tmp/target7371369939985176724arrow_orc_jni
#12 0x00007f613160f6cd in arrow::adapters::orc::ORCFileReader::Impl::Open(std::shared_ptr<arrow::io::RandomAccessFile> const&, arrow::MemoryPool*) ()
   from /tmp/target7371369939985176724arrow_orc_jni
#13 0x00007f6131609bd2 in arrow::adapters::orc::ORCFileReader::Open(std::shared_ptr<arrow::io::RandomAccessFile> const&, arrow::MemoryPool*) ()
   from /tmp/target7371369939985176724arrow_orc_jni
#14 0x00007f6131170270 in Java_org_apache_arrow_adapter_orc_OrcReaderJniWrapper_open () from /tmp/target7371369939985176724arrow_orc_jni
#15 0x00007f624fc6b9c0 in ?? ()
#16 0x0000000000000000 in ?? ()
```

lidavidm:
And interestingly now ORC passes in CI.

wgtmac:
From the backtrace, the ORC code does not look strange since it simply tried to call std::malloc for 2614 bytes...

lidavidm:
Yes, exactly. I can't figure out why it's failing sometimes.

That said it seems to be passing here (*knocks on wood*) so let's see...

lidavidm:
Nope, and now it fails again. It seems flaky or sensitive to something.

```
07:31:55.100 [main] DEBUG org.apache.orc.impl.OutStream -- Stream column 2 kind ROW_INDEX written to with Compress: ZLIB buffer: 262144
Fatal glibc error: malloc.c:2599 (sysmalloc): assertion failed: (old_top == initial_top (av) && old_size == 0) || ((unsigned long) (old_size) >= MINSIZE && prev_inuse (old_top) && ((unsigned long) old_end & (pagesize - 1)) == 0)
Aborted (core dumped)
```

wgtmac:
How can I reproduce it on my laptop?

lidavidm:
You can download the artifacts from the logs: https://github.com/apache/arrow-java/actions/runs/12543000171/job/34973510784?pr=449#step:8:33

(direct link: https://github.com/apache/arrow-java/actions/runs/12543000171/artifacts/2370586081)

Then extract to `arrow-java/dist`

```
lidavidm@lidavidm-nix ~/C/arrow-java (gh-13)> ls -R dist                                                                                                                               (base) 
dist:
arrow_cdata_jni/  arrow_dataset_jni/  arrow_orc_jni/  gandiva_jni/

dist/arrow_cdata_jni:
x86_64/

dist/arrow_cdata_jni/x86_64:
libarrow_cdata_jni.so*

...
```

Then run java_full_build.sh

```
$ ./ci/scripts/java_full_build.sh $(pwd) $(pwd)/arrow $(pwd)/dist
```

wgtmac:
The backtrace points to this line in the ORC codebase: https://github.com/apache/orc/blob/v2.0.3/c%2B%2B/src/Reader.cc#L1386

This is the first allocation (from the default memory pool) to read the entire file whose size is 2614.

lidavidm:
Hmm. It seems a simple malloc really shouldn't cause problems. And I don't think any of our wrapper code does anything nontrivial beforehand...

lidavidm:
Anyways, for now I've disabled the test so that the JNI build can pass. It turns out one of the fixes here is needed to fix the Arrow CI (https://github.com/apache/arrow/issues/45128).

ci/scripts/java_full_build.sh
kou, line 1: Can we remove `java_` prefix because this is the `apache/arrow-java` repository?

.github/workflows/test_jni.yml
```diff
@@ -0,0 +1,273 @@
...
+          - runs_on: ubuntu-latest
+            arch: "x86_64"
+            archery_arch: "amd64"
+            archery_arch_alias: "x86_64"
+            archery_arch_short: "amd64"
+    env:
+      # architecture name used for archery build
+      ARCH: ${{ matrix.platform.archery_arch }}
+      ARCH_ALIAS: ${{ matrix.platform.archery_arch_alias }}
+      ARCH_SHORT: ${{ matrix.platform.archery_arch_short }}
```
kou, line 56: We can remove them.

.github/workflows/test_jni.yml
```diff
@@ -0,0 +1,273 @@
...
+          path: arrow
+          submodules: recursive
+      - uses: docker/login-action@9780b0c442fbb1117ed29e0efdff1e18412f7567 # v3.3.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build C++ libraries
+        env:
+          VCPKG_BINARY_SOURCES: "clear;nuget,GitHub,readwrite"
```
kou, line 80: This may not work because we need more codes for this.
For example, https://github.com/apache/arrow/pull/44644/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R2135 and so on are needed.

But we can work on this as a follow-up task.

kou (approved):
+1

wgtmac:
I found that the JNI libraries built on ubuntu has linked with both jemalloc and mimalloc. The coredump indicates an invalid initial state in the sysmalloc. I'm not sure if it is an undefined behavior if we have enabled both jemalloc and mimalloc. Should we consider disable jemalloc by default?

ci/scripts/java_jni_manylinux_build.sh
```diff
@@ -0,0 +1,176 @@
...
+devtoolset_include_cpp="/opt/rh/devtoolset-${devtoolset_version}/root/usr/include/c++/${devtoolset_version}"
+: "${ARROW_ACERO:=ON}"
+export ARROW_ACERO
+: "${ARROW_BUILD_TESTS:=OFF}"
+: "${ARROW_DATASET:=ON}"
+export ARROW_DATASET
+: "${ARROW_GANDIVA:=ON}"
+export ARROW_GANDIVA
+: "${ARROW_GCS:=ON}"
+: "${ARROW_JEMALLOC:=ON}"
```
wgtmac, line 55:
```suggestion
: "${ARROW_JEMALLOC:=OFF}"
: "${ARROW_MIMALLOC:=ON}"
```

lidavidm:
I can try that. But I thought we've shipped multiple allocators in one binary before. (Arrow doesn't use jemalloc or mimalloc to replace system malloc.)

lidavidm:
Huh, that passed @wgtmac. Let's hope it stays that way :sweat_smile: 

I addressed Kou's feedback too.

.github/workflows/test_jni.yml
```diff
@@ -0,0 +1,267 @@
...
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build C++ libraries
+        run: |
+          docker compose run vcpkg-jni
+      - name: Push Docker image
+        if: success() && github.event_name == 'push' && github.repository == 'apache/arrow-java' && github.ref_name == 'main'
+        run: |
+          docker push ghcr.io/apache/arrow-java-dev:amd64-vcpkg-jni
```
kou, line 80:
```suggestion
          docker compose push vcpkg-jni
```

kou (approved):
+1

kou:
Hmm. The ORC crash is reproduced...?
https://github.com/apache/arrow-java/actions/runs/12557315120/job/35010522318?pr=449#step:7:19745

```text
Fatal glibc error: malloc.c:2599 (sysmalloc): assertion failed: (old_top == initial_top (av) && old_size == 0) || ((unsigned long) (old_size) >= MINSIZE && prev_inuse (old_top) && ((unsigned long) old_end & (pagesize - 1)) == 0)
Aborted (core dumped)
```

lidavidm:
I think it's flaky then. Are we ok with disabling it for now?

wgtmac:
Yes, we need to disable ORC test in this PR. Is it helpful to enable ASAN in a separate PR to build JNI libraries and use it for the ORC test?

lidavidm:
I can try again in another PR. I think I found that ASAN hid the problem (Valgrind too). (Also both tools are finicky with the JVM.)

lidavidm:
Filed https://github.com/apache/arrow-java/issues/473

lidavidm:
Now that we have some CI I'm going to start merging Dependabot updates again

wgtmac:
I tried to enable ASAN on Apache ORC and no issue has been found: https://github.com/apache/orc/actions/runs/12577822275/job/35055736586?pr=2097

lidavidm:
ASan replaces the malloc implementation, so the error may get masked. (Though if it is indeed memory corruption _presumably_ ASan would find that instead.)

lidavidm:
One other thing we could do is try various combinations of MALLOC_CHECK and MALLOC_PERTURB

wgtmac:
I still don't understand why we cannot see symbols of mimalloc from the coredump backtrace if we have linked mimalloc.

lidavidm:
We don't use mimalloc to replace malloc. It's only used by the Arrow memory pool. So we are still using glibc malloc for regular allocations.

PR #464, opened 2024-12-24
GH-463: Improve TZ support for JDBC driver

### LINKED ISSUE #463: [JDBC] Timezone improvements for the JDBC driver

Currently, our Timezone handling is pretty iffy as identified in a few different threads. I've been doing some digging and would like to propose a few changes. I will make a draft PR with how I think we should implement these changes so we can open up discussions on each point.

## Support `java.time` objects
The JDBC 4.2 spec specifies users can fetch `java.time` objects using `resultSet.getObject(..., <java.time>.class)`. We should support `OffsetDateTime`, `ZonedDateTime`, `Instant`, and `LocalDateTime` for timestamps. We should also support `LocalDate` for dates vectors, and `LocalTime` for times.

### Timestamp vectors that include a timezone should be treated as `TIMESTAMP_WITH_TIMEZONE`
- When retrieved as `OffsetDateTime` or `ZonedDateTime`, use the TZ included in the vector
- When retrieved as `Timestamp` with a Calendar, adjust the value to display time in the TZ specified by the Calendar, assuming the value in the vector is in the vector specified TZ
- When retrieved as `LocalDateTime`, do same as `Timestamp` but assume the Calendar specifies UTC.
- Still not sure how `Instant` plays into all of this

### Timestamp vectors without a timezone should be treated as `TIMESTAMP`
- When retrieved as `OffsetDateTime` or `ZonedDateTime`, throw an error?
- When retrieved as `Timestamp` or `LocalDateTime`, take the value as-is and don't do any TZ conversion. The assumption should be that if the vector didn't include TZ, we're just dealing with "wall-clock" time so it should never be adjusted
- Still not sure how `Instant` plays into all of this

### If we get a new vector type that includes TZ/Offset per record (https://github.com/apache/arrow/issues/44248), it should also be treated as `TIMESTAMP_WITH_TIMEZONE` 
- This is still a hypothetical, but I think if this comes to fruition we should treat it similar to Timestamp vectors with TZ, but use the TZ per record instead of for the whole vector.

## Don't do any TZ adjustments for regular `TIMESTAMP` w/o TZ
- We currently assume UTC for vectors w/o a TZ and still do an offset conversion when requested with a given Calendar. We should not do any TZ conversion for these since they should be treated as "wall-clock" time.

## Parameter binding

We probably also need to deal with relevant changes when binding parameters, but I'll leave that as a separate change since we also need to deal with #153.

Did I miss anything? Let me know!

aiguofer:
This PR adds support for natively fetching `java.time.*` objects through the JDBC driver.

DateVector
- getObject(LocalDate.class)

DateTimeVector
- getObject(OffsetDateTime.class)
- getObject(LocalDateTime.class)
- getObject(ZonedDateTime.class)
- getObject(Instant.class)

TimeVector
- getObject(LocalTime.class)

This PR also changes the behavior for vectors that include TZ info. These will now return as `TIMESTAMP_WITH_TIMEZONE`. 

The behavior for different ways to access a TimeStampVector are as follows:

|                                 | Vector with TZ                                                                                             | Vector W/O TZ                                                                             |
|---------------------------------|------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------|
| getTimestamp()                  | Get timestamp in the vector TZ                                                                             | Get timestamp in UTC                                                                      |
| getTimestamp(calendar)          | Get timestamp by adjusting from the vector TZ to the desired calendar TZ                                   | Get timestamp by adjusting from UTC to the desired calendar TZ (a bug, IMO)                            |
| getObject(LocalDateTime.class)  | Get LocalDateTime by taking the timestamp at the vector TZ and taking the "wall-clock" time at that moment | Treat the epoch value in the vector as the "wall-clock" time                              |
| getObject(Instant.class)        | Get Instant represented by the value in the vector TZ                                                      | Get Instant represented by the value in UTC                                               |
| getObject(OffsetDateTime.class) | Get OffsetDateTime represented by the value in the vector TZ (will account for daylight adjustment)        | Get OffsetDateTime represented by the value in UTC (will account for daylight adjustment) |
| getObject(ZonedDateTime.class)  | Get ZonedDateTime represented by the value in the vector at in its TZ                                      | Get ZonedDateTime represented by the value in the vector at in UTC                        |

Closes #463

### DIFF
--- a/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/accessor/impl/calendar/ArrowFlightJdbcTimeStampVectorAccessor.java
+++ b/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/accessor/impl/calendar/ArrowFlightJdbcTimeStampVectorAccessor.java
@@ -160,7 +160,7 @@ public class ArrowFlightJdbcTimeStampVectorAccessor extends ArrowFlightJdbcAcces
       return null;
     }
 
-    return new Date(Timestamp.valueOf(localDateTime).getTime());
+    return new Date(getTimstampWithOffset(calendar, localDateTime).getTime());
   }
 
   @Override
@@ -170,7 +170,7 @@ public class ArrowFlightJdbcTimeStampVectorAccessor extends ArrowFlightJdbcAcces
       return null;
     }
 
-    return new Time(Timestamp.valueOf(localDateTime).getTime());
+    return new Time(getTimstampWithOffset(calendar, localDateTime).getTime());
   }
 
   @Override
@@ -179,19 +179,26 @@ public class ArrowFlightJdbcTimeStampVectorAccessor extends ArrowFlightJdbcAcces
     if (localDateTime == null) {
       return null;
     }
-    // to prevent breaking changes for those using this, apply the offset that was previously
-    // applied in getLocalDateTime
+
+    return getTimstampWithOffset(calendar, localDateTime);
+  }
+
+  /**
+   * Apply offset to LocalDateTime to get a Timestamp with legacy behavior. Previously we applied
+   * the offset to the LocalDateTime even if the underlying Vector did not have a TZ. In order to
+   * support java.time.* accessors, we fixed this so we only apply the offset if the underlying
+   * vector includes TZ info. In order to maintain backward compatibility, we apply the offset if
+   * needed for getDate, getTime, and getTimestamp.
+   */
+  private Timestamp getTimstampWithOffset(Calendar calendar, LocalDateTime localDateTime) {
     if (calendar != null && !isZoned) {
       TimeZone timeZone = calendar.getTimeZone();
       long millis = Timestamp.valueOf(localDateTime).getTime();
       localDateTime =
           localDateTime.minus(
               timeZone.getOffset(millis) - this.timeZone.getOffset(millis), ChronoUnit.MILLIS);
-
-      return Timestamp.valueOf(localDateTime);
-    } else {
-      return Timestamp.valueOf(localDateTime);
     }
+    return Timestamp.valueOf(localDateTime);
   }
 
   protected static TimeUnit getTimeUnitForVector(TimeStampVector vector) {
--- a/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/accessor/impl/calendar/ArrowFlightJdbcTimeStampVectorAccessor.java
+++ b/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/accessor/impl/calendar/ArrowFlightJdbcTimeStampVectorAccessor.java
@@ -82,8 +82,10 @@ public class ArrowFlightJdbcTimeStampVectorAccessor extends ArrowFlightJdbcAcces
   @Override
   public <T> T getObject(final Class<T> type) throws SQLException {
     final Object value;
-    if (!this.isZoned & Set.of(OffsetDateTime.class, ZonedDateTime.class, Instant.class).contains(type)) {
-      throw new SQLException("Vectors without timezones can't be converted to objects with offset/tz info.");
+    if (!this.isZoned
+        & Set.of(OffsetDateTime.class, ZonedDateTime.class, Instant.class).contains(type)) {
+      throw new SQLException(
+          "Vectors without timezones can't be converted to objects with offset/tz info.");
     } else if (type == OffsetDateTime.class) {
       value = getOffsetDateTime();
     } else if (type == LocalDateTime.class) {
--- a/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/accessor/impl/calendar/ArrowFlightJdbcTimeStampVectorAccessor.java
+++ b/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/accessor/impl/calendar/ArrowFlightJdbcTimeStampVectorAccessor.java
@@ -31,6 +31,7 @@ import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.time.temporal.ChronoUnit;
 import java.util.Calendar;
+import java.util.Objects;
 import java.util.TimeZone;
 import java.util.concurrent.TimeUnit;
 import java.util.function.IntSupplier;
@@ -71,7 +72,9 @@ public class ArrowFlightJdbcTimeStampVectorAccessor extends ArrowFlightJdbcAcces
     this.holder = new Holder();
     this.getter = createGetter(vector);
 
+    // whether the vector included TZ info
     this.isZoned = getVectorIsZoned(vector);
+    // non-null, either the vector TZ or default to UTC
     this.timeZone = getTimeZoneForVector(vector);
     this.timeUnit = getTimeUnitForVector(vector);
     this.longToLocalDateTime = getLongToLocalDateTimeForVector(vector, this.timeZone);
@@ -231,11 +234,7 @@ public class ArrowFlightJdbcTimeStampVectorAccessor extends ArrowFlightJdbcAcces
     ArrowType.Timestamp arrowType =
         (ArrowType.Timestamp) vector.getField().getFieldType().getType();
 
-    String timezoneName = arrowType.getTimezone();
-    if (timezoneName == null) {
-      return TimeZone.getTimeZone("UTC");
-    }
-
+    String timezoneName = Objects.requireNonNullElse(arrowType.getTimezone(), "UTC");
     return TimeZone.getTimeZone(timezoneName);
   }
 
--- a/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/accessor/impl/calendar/ArrowFlightJdbcTimeStampVectorAccessor.java
+++ b/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/accessor/impl/calendar/ArrowFlightJdbcTimeStampVectorAccessor.java
@@ -35,16 +35,13 @@ import java.util.Objects;
 import java.util.TimeZone;
 import java.util.concurrent.TimeUnit;
 import java.util.function.IntSupplier;
-
 import org.apache.arrow.driver.jdbc.accessor.ArrowFlightJdbcAccessor;
 import org.apache.arrow.driver.jdbc.accessor.ArrowFlightJdbcAccessorFactory;
 import org.apache.arrow.vector.TimeStampVector;
 import org.apache.arrow.vector.types.pojo.ArrowType;
 import org.apache.arrow.vector.util.DateUtility;
 
-/**
- * Accessor for the Arrow types extending from {@link TimeStampVector}.
- */
+/** Accessor for the Arrow types extending from {@link TimeStampVector}. */
 public class ArrowFlightJdbcTimeStampVectorAccessor extends ArrowFlightJdbcAccessor {
 
   private final TimeZone timeZone;
@@ -54,16 +51,12 @@ public class ArrowFlightJdbcTimeStampVectorAccessor extends ArrowFlightJdbcAcces
   private final Holder holder;
   private final boolean isZoned;
 
-  /**
-   * Functional interface used to convert a number (in any time resolution) to LocalDateTime.
-   */
+  /** Functional interface used to convert a number (in any time resolution) to LocalDateTime. */
   interface LongToLocalDateTime {
     LocalDateTime fromLong(long value);
   }
 
-  /**
-   * Instantiate a ArrowFlightJdbcTimeStampVectorAccessor for given vector.
-   */
+  /** Instantiate a ArrowFlightJdbcTimeStampVectorAccessor for given vector. */
   public ArrowFlightJdbcTimeStampVectorAccessor(
       TimeStampVector vector,
       IntSupplier currentRowSupplier,
@@ -148,7 +141,8 @@ public class ArrowFlightJdbcTimeStampVectorAccessor extends ArrowFlightJdbcAcces
 
     LocalDateTime localDateTime = this.longToLocalDateTime.fromLong(value);
 
-    // Adjust timestamp to desired calendar (if provided) only if the column includes TZ info, otherwise treat as wall-clock time
+    // Adjust timestamp to desired calendar (if provided) only if the column includes TZ info,
+    // otherwise treat as wall-clock time
     if (calendar != null && this.isZoned) {
       TimeZone timeZone = calendar.getTimeZone();
       long millis = this.timeUnit.toMillis(value);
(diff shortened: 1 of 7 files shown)


flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/accessor/impl/calendar/ArrowFlightJdbcTimeStampVectorAccessor.java
```diff
@@ -58,6 +71,7 @@ public ArrowFlightJdbcTimeStampVectorAccessor(
     this.holder = new Holder();
     this.getter = createGetter(vector);
 
+    this.isZoned = getVectorIsZoned(vector);
     this.timeZone = getTimeZoneForVector(vector);
```
jduo, line 75: It's a bit strange that both the timeZone and isZoned field exist and would be good to clarify why/how they are used and are different.

flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/utils/SqlTypes.java
```diff
@@ -120,7 +120,12 @@ public static int getSqlTypeIdFromArrowType(ArrowType arrowType) {
       case Time:
         return Types.TIME;
       case Timestamp:
-        return Types.TIMESTAMP;
+        String tz = ((ArrowType.Timestamp) arrowType).getTimezone();
+        if (tz != null){
```
jduo, line 124: This makes sense.

flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/accessor/impl/calendar/ArrowFlightJdbcTimeStampVectorAccessor.java
aiguofer: Yeah, originally I thought about setting tz to null if not available but there's some expectations about `timeZone` not being null and having some default value in a few places where it felt like untangling would make this too big. I'll add some comments.

linghengqian (commented):
- I tested this PR from https://github.com/apache/arrow-java/issues/636 and https://github.com/influxdata/influxdb/issues/25983 , but from my point of view, the current PR still handles `java.time.Instant` incorrectly.

- The `java.time.Instant` obtained from the query is 8 hours different from the original timestamp, which is still unreasonable. See https://github.com/linghengqian/influxdb-3-core-jdbc-test/commit/8ae1aaa2fd0d715b4dbe40ca28d6f64d6b918375 .

```shell
sdk install java 21.0.6-ms
sdk use java 21.0.6-ms

git clone git@github.com:aiguofer/arrow-java.git -b improved_tz_support
cd ./arrow/
mvn clean install -DskipTests -Dspotless.check.skip=true
cd ../

git clone git@github.com:linghengqian/influxdb-3-core-jdbc-test.git
cd ./influxdb-3-core-jdbc-test/
sdk use java 21.0.6-ms
./mvnw -T 1C -Dtest=TimeDifferenceTest clean test
```
<details>
<summary>Click me to view the error log of the unit test🥯🥨🍟🧂🥖🥚🍔🦪🍜🍘</summary>

```shell
$ ./mvnw -T 1C -Dtest=TimeDifferenceTest clean test
[INFO] Scanning for projects...
[INFO] 
[INFO] Using the MultiThreadedBuilder implementation with a thread count of 16
[INFO] 
[INFO] ----------< io.github.linghengqian:influxdb-3-core-jdbc-test >----------
[INFO] Building influxdb-3-core-jdbc-test 1.0-SNAPSHOT
[INFO]   from pom.xml
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] 
[INFO] --- clean:3.2.0:clean (default-clean) @ influxdb-3-core-jdbc-test ---
[INFO] Deleting /home/linghengqian/TwinklingLiftWorks/git/public/influxdb-3-core-jdbc-test/target
[INFO] 
[INFO] --- resources:3.3.1:resources (default-resources) @ influxdb-3-core-jdbc-test ---
[INFO] skip non existing resourceDirectory /home/linghengqian/TwinklingLiftWorks/git/public/influxdb-3-core-jdbc-test/src/main/resources
[INFO] 
[INFO] --- compiler:3.13.0:compile (default-compile) @ influxdb-3-core-jdbc-test ---
[INFO] No sources to compile
[INFO] 
[INFO] --- resources:3.3.1:testResources (default-testResources) @ influxdb-3-core-jdbc-test ---
[INFO] skip non existing resourceDirectory /home/linghengqian/TwinklingLiftWorks/git/public/influxdb-3-core-jdbc-test/src/test/resources
[INFO] 
[INFO] --- compiler:3.13.0:testCompile (default-testCompile) @ influxdb-3-core-jdbc-test ---
[INFO] Recompiling the module because of changed source code.
[INFO] Compiling 7 source files with javac [debug target 21] to target/test-classes
[INFO] 
[INFO] --- surefire:3.5.2:test (default-test) @ influxdb-3-core-jdbc-test ---
[INFO] Using auto detected provider org.apache.maven.surefire.junitplatform.JUnitPlatformProvider
[INFO] 
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running io.github.linghengqian.TimeDifferenceTest
SLF4J(W): No SLF4J providers were found.
SLF4J(W): Defaulting to no-operation (NOP) logger implementation
SLF4J(W): See https://www.slf4j.org/codes.html#noProviders for further details.
2月 25, 2025 11:11:36 上午 org.apache.arrow.driver.jdbc.shaded.org.apache.arrow.memory.BaseAllocator <clinit>
信息: Debug mode disabled. Enable with the VM option -Darrow.memory.debug.allocator=true.
2月 25, 2025 11:11:36 上午 org.apache.arrow.driver.jdbc.shaded.org.apache.arrow.memory.DefaultAllocationManagerOption getDefaultAllocationManagerFactory
信息: allocation manager type not specified, using netty as the default type
2月 25, 2025 11:11:36 上午 org.apache.arrow.driver.jdbc.shaded.org.apache.arrow.memory.CheckAllocator reportResult
信息: Using DefaultAllocationManager at memory/netty/DefaultAllocationManagerFactory.class
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 3.419 s <<< FAILURE! -- in io.github.linghengqian.TimeDifferenceTest
[ERROR] io.github.linghengqian.TimeDifferenceTest.test -- Time elapsed: 3.354 s <<< FAILURE!
java.lang.AssertionError: 

Expected: is <2025-02-25T03:11:24.614425710Z>
     but: was <2025-02-24T19:11:24.614425710Z>
        at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
        at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:8)
        at io.github.linghengqian.TimeDifferenceTest.queryDataByJdbcDriver(TimeDifferenceTest.java:89)
        at io.github.linghengqian.TimeDifferenceTest.test(TimeDifferenceTest.java:50)
        at java.base/java.lang.reflect.Method.invoke(Method.java:580)
        at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
        at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   TimeDifferenceTest.test:50->queryDataByJdbcDriver:89 
Expected: is <2025-02-25T03:11:24.614425710Z>
     but: was <2025-02-24T19:11:24.614425710Z>
[INFO] 
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  4.894 s (Wall Clock)
[INFO] Finished at: 2025-02-25T11:11:38+08:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.5.2:test (default-test) on project influxdb-3-core-jdbc-test: There are test failures.
[ERROR] 
[ERROR] See /home/linghengqian/TwinklingLiftWorks/git/public/influxdb-3-core-jdbc-test/target/surefire-reports for the individual test results.
[ERROR] See dump files (if any exist) [date].dump, [date]-jvmRun[N].dump and [date].dumpstream.
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException
```
</details>

- Another issue that is not resolved by the current PR is that the file is not formatted, resulting in the need for a command line argument of `-Dspotless.check.skip=true`.

aiguofer:
@linghengqian I looked at your test but there's a few things that are not immediately clear.

- Does Influx adjust the TZ for the Point based on its location? (that is, does it convert `magicTime` to London TZ?)
- Which timestamp vector does it use for transport when querying `time`? does it include TZ info or not?

lidavidm:
Hopefully/presumably the `Instant` value should be correct, though? Although I see it's an `Instant` obtained from the `java.sql.Timestamp` which is already suspect (it's never been clear to me what the expectation is for the semantics of that class)

aiguofer:
Awesome! `resultSet.getObject("time", Instant.class)` is the right way to do it so I'm glad that's working. I'd like to also add some unit tests before finishing the PR but I'm pretty busy for the next month and a half.

I can easily make the formatting fixes for now and remove the draft status to get some more reviews. Depending on how others feel about unit tests, it would be great if someone can help contribute the tests.

aiguofer:
Ok I tried to fix the existing tests but now I'm worried that the change to not modify the offset for Timestamp vectors w/o TZ info might have an inconsistent behavior with `TimeVector` accessors, `DateVector` accessors,  and `getTimestamp` accessor for `VarCharVector`.

I also noticed that what we currently do for `TimeStampVector` is already inconsistent with what we do for `TimeVector` and `DateVector` accessors. For the latter, we use `DateTimeUtils.applyCalendarOffset`, which uses `TimeZone.getDefault()` to determine our current timezone. However, for `TimeStampAccessor` we assume `UTC` if not provided.

There's a lot of inter-related parts here and breaking changes could be troublesome.

I'm going to change behavior so that usage of the "legacy" JDBC date/time objects stays the same and we can simply recommend that users use `getObject(<java.time.* class>)` to get expected behaviors.

lidavidm:
I'll take a look. That sounds reasonable as a first step but we should probably fix things longer term if we can. 

@laurentgo not sure if you have ideas on how the native JDBC datetime types are supposed to behave, I found the documentation rather under-specified...

flight/flight-sql-jdbc-core/src/test/java/org/apache/arrow/driver/jdbc/accessor/impl/calendar/ArrowFlightJdbcTimeStampVectorAccessorTest.java
```diff
@@ -320,6 +404,24 @@ private Timestamp getTimestampForVector(int currentRow, String timeZone) {
...
+    ZonedDateTime expectedTimestamp = null;
+    if (object instanceof LocalDateTime) {
+      expectedTimestamp = ((LocalDateTime) object).atZone(tz.toZoneId());
+    } else if (object instanceof Long) {
+      TimeUnit timeUnit = getTimeUnitForVector(vector);
+      long millis = timeUnit.toMillis((Long) object);
+      long offset = tz.getOffset(millis);
+      // TODO: should we actually add the offset here? I'm not completely sure how the value is
+      // stored in the vector
+      LocalDateTime local = new Timestamp(millis + offset).toLocalDateTime();
```
aiguofer, line 419: I'm not entirely sure about this part. I'm assuming that the Long stored in the vector is the Epoch value, so we do need to adjust for the given TZ.

aiguofer:
Awesome thanks! I updated a few more things including a table in the PR description showing different behaviors for `TimeStampVectors`. We might want to cover all scenarios and add documentation.

I'm not sure how much more I'll be able to work on this over the next few weeks but I'll try to keep up with it!

flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/accessor/impl/calendar/ArrowFlightJdbcTimeStampVectorAccessor.java
```diff
@@ -68,11 +78,57 @@ public Class<?> getObjectClass() {
...
+    }
+    return !type.isPrimitive() && wasNull ? null : type.cast(value);
+  }
+
   @Override
   public Object getObject() {
     return this.getTimestamp(null);
   }
 
+  private ZonedDateTime getZonedDateTime() {
```
lidavidm, line 105: I think we shouldn't allow getting these objects (Instant, ZonedDateTime, OffsetDateTime) unless we know the underlying vector is also zoned. Otherwise we're just lying to the user.

lidavidm (commented):
Looking through the implementation overall I'm starting to doubt its correctness. In particular both zoned and unzoned timestamps are treated via LocalDateTime, but I think this is wrong for timestamps with a timezone: as I understand, the conversion from Instant to LocalDateTime is not unambiguous (if we have a named time zone and not an offset, a wall clock time may correspond to different instants around a daylight savings time transition). I'm also still doubtful about how we are treating the `Calendar` parameter; the JDBC documentation generally says

```
This method uses the given calendar to construct an appropriate millisecond value for the timestamp if the underlying database does not store timezone information.
```

Which to me means that we should only use the Calendar when the column does _not_ include TZ info.

lidavidm:
The ambiguity in the JDBC documentation is frustrating, though; I really can't tell what is "correct" or not.

lidavidm:
FWIW, I've been working on cataloguing what JDBC drivers do here and trying to compare it to the current Flight SQL JDBC behavior + I will compare to this PR too

flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/accessor/impl/calendar/ArrowFlightJdbcTimeStampVectorAccessor.java
```diff
@@ -102,7 +160,7 @@ public Date getDate(Calendar calendar) {
       return null;
     }
 
-    return new Date(Timestamp.valueOf(localDateTime).getTime());
+    return new Date(getTimstampWithOffset(calendar, localDateTime).getTime());
```
lidavidm, line 163: typo: `getTimestamp` not `getTimstamp`

lidavidm (commented):
Ok. I think let's merge this PR, but not letting you get Instant/etc. unless the underlying type is zoned. And I think the handling of timestamps in general needs to be overhauled, but this is mostly orthogonal to that.

aiguofer:
@lidavidm sorry been super busy, but thanks for the feedback! I'll fix that typo and throw an exception if trying to fetch objects with offset info when the underlying vector does not.

I agree we probably want to overhaul the timestamp handling in general as the current behavior is incorrect IMO, but that can be done separately.

flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/accessor/impl/calendar/ArrowFlightJdbcTimeStampVectorAccessor.java
aiguofer: done

lidavidm (approved):

lidavidm:
It seems tests needed to be adjusted too

aiguofer:
AHh sorry, forgot about ensuring tests passed :( tests fixed now!

PR #467, opened 2024-12-26
GH-466: Export namespace from Flight package

### LINKED ISSUE #466: Modules aren't properly declared

- org.apache.arrow.flight.grpc isn't exported
- JDBC driver uses 'flight.sql.jdbc.core' which doesn't match other packages like 'org.apache.arrow.flight.core'

lidavidm:
Fixes #466.


kou (approved):
+1

wgtmac (approved):

PR #488, opened 2025-01-07
allow to override Base(Large)RepeatedValueViewVector

rluvaton:
This is done as different arrow implementation have different default list item field name

For example, Rust implementation of Arrow (`arrow-rs`) has default value of `item` (https://github.com/apache/arrow-rs/blob/4f1f6e57c568fae8233ab9da7d7c7acdaea4112a/arrow-schema/src/field.rs#L125).

And according to the specification, the name of the list is not required to be a specific string. so to make it easier to use Java Arrow implementation with other implementation of Arrow, I'm making the `DATA_VECTOR_NAME` configurable

### DIFF
--- a/vector/src/main/java/org/apache/arrow/vector/complex/BaseLargeRepeatedValueViewVector.java
+++ b/vector/src/main/java/org/apache/arrow/vector/complex/BaseLargeRepeatedValueViewVector.java
@@ -44,7 +44,7 @@ import org.apache.arrow.vector.util.SchemaChangeRuntimeException;
 public abstract class BaseLargeRepeatedValueViewVector extends BaseValueVector
     implements RepeatedValueVector, FieldVector {
   public static final FieldVector DEFAULT_DATA_VECTOR = ZeroVector.INSTANCE;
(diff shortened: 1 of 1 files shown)


lidavidm (commented):
I think it'd make more sense to allow changing `defaultDataVectorName` for an instance? I'm not a big fan of mutable globals

rluvaton:
> I think it'd make more sense to allow changing `defaultDataVectorName` for an instance? I'm not a big fan of mutable globals

Changing `defaultDataVectorName` for an instance is too late, no? as the instance already created

lidavidm:
The child vector is created lazily, though

rluvaton:
But it can still be too late, is there a way to change a field (or field name) of existing list?

lidavidm:
I don't see how a mutable global lets you change the name of an existing list either, though.

Once the child is created I think you'd have to recreate the child instead of being able to mutate the name.

lidavidm:
Do you have a sample to show the scenario you are concerned about?

Issue #492, opened 2025-01-08
[CI] Remove `java_` prefix from `ci/scripts/*.sh`


kou:
### Describe the enhancement requested

Because this repository contains only the Java implementation.

PR #493, opened 2025-01-08
GH-492: [CI] Remove `java_` prefix from `ci/scripts/*.sh`

kou:
Fixes GH-492.


kou:
TODO: We need to update https://github.com/apache/arrow/blob/438cf9bd2d31506635bc56ca0371df53030ff4d1/ci/scripts/integration_arrow_build.sh#L63 after this is merged.

lidavidm (approved):

PR #495, opened 2025-01-09
GH-494: [Flight] Allow configuring connect timeout in JDBC

### LINKED ISSUE #494: [Flight] JDBC driver: make connect timeout adjustible

- The default connect timeout is 30 seconds. This is rather painful when the server returns an address that isn't reachable. (Example: the server is deployed on a Kubernetes cluster that is being accessed via proxy/port-forwarding, and the server has a separated data plane/control plane.)

lidavidm:
Allow configuring the connect timeout via `connectTimeoutMs` so that the driver doesn't wait so long before giving up on unreachable locations.

Fixes #494.

### DIFF
--- a/flight/flight-sql-jdbc-core/pom.xml
+++ b/flight/flight-sql-jdbc-core/pom.xml
@@ -48,6 +48,21 @@ under the License.
     </dependency>
 
     <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-api</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-netty</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>io.netty</groupId>
+      <artifactId>netty-transport</artifactId>
+    </dependency>
+
+    <dependency>
       <groupId>org.apache.arrow</groupId>
       <artifactId>arrow-memory-core</artifactId>
     </dependency>
--- a/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightConnection.java
+++ b/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightConnection.java
@@ -113,6 +113,7 @@ public final class ArrowFlightConnection extends AvaticaConnection {
           .withRetainCookies(config.retainCookies())
           .withRetainAuth(config.retainAuth())
           .withCatalog(config.getCatalog())
+          .withConnectTimeout(config.getConnectTimeout())
           .build();
     } catch (final SQLException e) {
       try {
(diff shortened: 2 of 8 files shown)


lidavidm:
@jbonofre  are you familiar with the JDBC driver at all?

jbonofre:
@lidavidm yes a bit. I can take a look on this one.

flight/flight-core/src/main/java/org/apache/arrow/flight/grpc/NettyClientBuilder.java
```diff
@@ -0,0 +1,232 @@
...
+  public NettyClientBuilder verifyServer(boolean verifyServer) {
+    this.verifyServer = verifyServer;
+    return this;
+  }
+
+  /** Create the client from this builder. */
+  public NettyChannelBuilder build() {
+    final NettyChannelBuilder builder;
+
+    switch (location.getUri().getScheme()) {
```
jbonofre, line 137: FYI, I'm investigating the Netty channel address depending of the scheme as it seems we have a weird behavior here.
Your change move that from `FlightClient` to `NettyClientBuilder` so it looks good to me.

flight/flight-sql-jdbc-core/pom.xml
```diff
@@ -47,6 +47,21 @@ under the License.
       </exclusions>
     </dependency>
 
+    <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-api</artifactId>
```
jbonofre, line 52: I'm surprised we have to explicitly define these dependencies, I thought they came transitively from arrow flight.

flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightConnection.java
```diff
@@ -113,6 +114,8 @@ private static ArrowFlightSqlClientHandler createNewClientHandler(
           .withRetainCookies(config.retainCookies())
           .withRetainAuth(config.retainAuth())
           .withCatalog(config.getCatalog())
+          .withClientCache(config.useClientCache() ? new FlightClientCache() : null)
```
jbonofre, line 117: That's a new feature (client caching), right ? It's ok in the PR but it looks a big "unrelated" to the PR title 😄

jbonofre (approved):
It looks good to me.

My only comment is that I would expect client cache in another PR as it's a new feature compared to the initial intend of this PR.

lidavidm:
Thanks, I can split up this PR into 3? parts if that would be better.

jbonofre:
@lidavidm no, that's OK. I was just a bit confused because the change seems larger than just the PR title. But it's OK, I'm fine with that. I will do a new review pass. Thanks !

lidavidm:
It's probably better for the changelog if I split it up so no worries! Thanks!

lidavidm:
Ok, I've cut this down to _just_ the connect timeout option. I'll follow up with a third PR to add back the basic client cache. (There's yet another issue for adding a full client cache.)

lidavidm:
Going to merge since there's been no further comments.

Issue #20, opened 2024-11-25
Add release CI


kou:
See also: https://github.com/apache/arrow-go/pull/122

kou:
I'm working on this.

PR #496, opened 2025-01-09
GH-20: Add release scripts

kou:
Fixes GH-20.

This is based on https://github.com/apache/arrow-go/pull/122 .

Workflow:

Cut a RC:

1. Bump version by `dev/release/bump_version.sh 19.0.0`
2. Run `dev/release/release_rc.sh`
   1. `dev/release/release_rc.sh` pushes `v${version}-rc${rc}` tag
   2. `.github/workflows/rc.yml` creates `apache-arrow-java-${version}.tar.gz{,.sha256,.sha512}`
   3. `.github/workflows/rc.yml` uploads `apache-arrow-java-${version}.tar.gz{,.sha256,.sha512}` to GitHub Releases
   4. `dev/release/release_rc.sh` downloads `apache-arrow-java-${version}.tar.gz` from GitHub Releases
   5. `dev/release/release_rc.sh` signs `apache-arrow-java-${version}.tar.gz` as `apache-arrow-go-${version}.tar.gz.asc`
   6. `dev/release/release_rc.sh` uploads `apache-arrow-java-${version}.tar.gz.asc` to GitHub Releases
3. Start a vote

(GitHub Actions instead of
https://dist.apache.org/repos/dist/dev/arrow/ is used like ADBC.)

Verify a RC:

1. Run `dev/release/verify_rc.sh`

Release an approved RC:

1. Run `dev/release/release.sh`
   1. `dev/release/release.sh` pushes `v${version}` tag that refers that same commit ID as `v${version}-rc${rc}`
   2. `.github/workflows/release.yml` create a GitHub Releases for `v${version}` tag by copying the RC GitHub Releases
   3. `dev/release/release.sh` downloads `apache-arrow-java-${version}.tar.gz{,.asc,.sha256,.sha512}` from GitHub Releases
   4. `dev/release/release.sh` uploads `apache-arrow-java-${version}.tar.gz{,.asc,.sha256,.sha512}` to https://dist.apache.org/repos/dist/release/arrow
   5. `dev/release/release.sh` removes old releases from https://dist.apache.org/repos/dist/release/arrow
2. Add this release to ASF's report database: https://reporter.apache.org/addrelease.html?arrow

Follow-up tasks:
* Add support for building binary packages (.jar)
* Add support for releasing binary packages (.jar)
* Add support for running JNI test in the verification script
* Add support for running integration test in the verification script
* Add support for verifying binary packages (.jar) in the verification script

### DIFF
--- a/.github/workflows/rc.yml
+++ b/.github/workflows/rc.yml
@@ -106,8 +106,6 @@ jobs:
           VERIFY_DEFAULT=0 \
             VERIFY_SOURCE=1 \
             dev/release/verify_rc.sh "${version}" "${rc}"
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   upload:
     name: Upload
     if: github.ref_type == 'tag'
--- a/.github/workflows/rc.yml
+++ b/.github/workflows/rc.yml
@@ -1,0 +1,141 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+name: RC
+on:
+  push:
+    branches:
+      - '**'
+      - '!dependabot/**'
+    tags:
+      - '*-rc*'
+  pull_request:
+concurrency:
+  group: ${{ github.repository }}-${{ github.head_ref || github.sha }}-${{ github.workflow }}
+  cancel-in-progress: true
+permissions:
+  contents: read
+jobs:
+  archive:
+    name: Archive
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          submodules: recursive
+      - name: Prepare for tag
+        if: github.ref_type == 'tag'
+        run: |
+          version=${GITHUB_REF_NAME%-rc*}
+          version=${version#v}
+          rc=${GITHUB_REF_NAME#*-rc}
+          echo "VERSION=${version}" >> ${GITHUB_ENV}
+          echo "RC=${rc}" >> ${GITHUB_ENV}
+      - name: Prepare for branch
+        if: github.ref_type == 'branch'
+        run: |
+          version=$(grep -o '^  <version>.*</version>' "pom.xml" |
+            sed \
+              -e 's,^  <version>,,' \
+              -e 's,</version>$,,')
+          rc=$(date +%Y%m%d)
+          echo "VERSION=${version}" >> ${GITHUB_ENV}
+          echo "RC=${rc}" >> ${GITHUB_ENV}
+      - name: Archive
+        run: |
+          id="apache-arrow-java-${VERSION}"
+          tar_gz="${id}.tar.gz"
+          echo "TAR_GZ=${tar_gz}" >> ${GITHUB_ENV}
+          git archive HEAD --prefix "${id}/" --output "${tar_gz}"
+          sha256sum "${tar_gz}" > "${tar_gz}.sha256"
+          sha512sum "${tar_gz}" > "${tar_gz}.sha512"
+      - name: Audit
+        run: |
+          dev/release/run_rat.sh "${TAR_GZ}"
+      - uses: actions/upload-artifact@6f51ac03b9356f520e9adb1b1b7802705f340c2b # v4.5.0
+        with:
+          name: archive
+          path: |
+            apache-arrow-java-*
+  verify:
+    name: Verify
+    needs:
+      - archive
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - macos-latest
+          - ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          submodules: recursive
+      - uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16 # v4.1.8
+        with:
+          name: archive
+      - name: Verify
+        run: |
+          tar_gz=$(echo apache-arrow-java-*.tar.gz)
+          version=${tar_gz#apache-arrow-java-}
+          version=${version%.tar.gz}
+          # rc isn't used with VERIFY_DOWNLOAD=0
+          if [ "${GITHUB_REF_TYPE}" = "tag" ]; then
+            rc="${GITHUB_REF_NAME#*-rc}"
+          else
+            rc=$(date +%Y%m%d)
+          fi
+          VERIFY_DEFAULT=0 \
+            VERIFY_SOURCE=1 \
+            dev/release/verify_rc.sh "${version}" "${rc}"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  upload:
+    name: Upload
+    if: github.ref_type == 'tag'
+    needs:
+      - verify
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          submodules: recursive
+      - uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16 # v4.1.8
+        with:
+          name: archive
+      - name: Upload
+        run: |
+          # TODO: Add support for release note
+          version=${GITHUB_REF_NAME%-rc*}
+          version=${version#v}
+          rc=${GITHUB_REF_NAME#*-rc}
+          gh release create ${GITHUB_REF_NAME} \
+            --notes "TODO" \
+            --prerelease \
+            --title "Apache Arrow Java ${version} RC${rc}" \
+            --verify-tag \
+            apache-arrow-java-*.tar.gz \
+            apache-arrow-java-*.tar.gz.sha*
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/rc.yml
+++ b/.github/workflows/rc.yml
@@ -124,12 +124,12 @@ jobs:
           name: archive
       - name: Upload
         run: |
-          # TODO: Add support for release note
+          # GH-499: How to create release notes?
           version=${GITHUB_REF_NAME%-rc*}
           version=${version#v}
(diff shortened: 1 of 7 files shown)


kou:
I'll test this on my fork.
If you have any concerns for this workflow, please share them.

dev/release/release_rc.sh
```diff
@@ -0,0 +1,152 @@
...
+    echo "This script must be ran with working copy of apache/arrow-java."
+    echo "The origin's URL: ${git_origin_url}"
+    exit 1
+  fi
+fi
+
+if [ "${RELEASE_PULL}" -gt 0 ]; then
+  echo "Ensure using the latest commit"
+  git checkout main
+  git pull --rebase --prune
```
lidavidm, line 53: Hmm, wouldn't `--ff-only` be better in case there's an unpushed local commit for some reason?

.github/workflows/rc.yml
```diff
@@ -0,0 +1,141 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+name: RC
+on:
```
lidavidm, line 19: This might also make sense to run nightly to test the process? (We could perhaps upload to a prerelease and use that as our 'nightly artifacts'?)

lidavidm (approved):

dev/release/README.md
```diff
@@ -0,0 +1,122 @@
...
+  specific language governing permissions and limitations
+  under the License.
+-->
+
+# Release
+
+## Overview
+
+  1. Test the revision to be released
+  2. Bump version by `dev/release/bump_version.sh X.Y.Z`
```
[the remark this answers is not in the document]
kou: It makes sense.
I'll do it.

.github/workflows/release.yml
```diff
@@ -0,0 +1,53 @@
...
+          set -x
+          latest_rc_tag=$(gh release list --json tagName --jq '.[].tagName' | \
+                            grep -F "${GITHUB_REF_NAME}-rc" | \
+                            head -n1)
+          gh release download ${latest_rc_tag} --dir dists
+      - name: Create GitHub Release
+        run: |
+          version=${GITHUB_REF_NAME#v}
+          gh release create ${GITHUB_REF_NAME} \
+            --notes "TODO" \
```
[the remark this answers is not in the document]
kou: Let's discuss this as a separated task: #499

.github/workflows/rc.yml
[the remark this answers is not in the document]
kou: Ah, it's a good idea!
Can I merge the existing `test_jni.yml` to here?
Or should we keep these workflows separated?

.github/workflows/rc.yml
kou: It's a good idea!
I'll add `schedule:`.

dev/release/release_rc.sh
kou: Good catch! We should use `--ff-only` here. I'll fix this.

.github/workflows/rc.yml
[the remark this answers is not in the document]
lidavidm: I don't think we need to tackle this right away. GitHub lets you run actions as a substep, so maybe we can run test_jni.yml as a substep, then pull the artifacts back into the release pipeline.

dev/release/bump_version.sh
```diff
@@ -0,0 +1,51 @@
...
+mvn versions:set "-DnewVersion=${version}" -DprocessAllModules -DgenerateBackupPoms=false
+case "${version}" in
+*-SNAPSHOT)
+  tag=main
+  ;;
+*)
+  tag=v${version}
+  ;;
+esac
+mvn versions:set-scm-tag "-DnewTag=${tag}" -DgenerateBackupPoms=false -pl :arrow-java-root
```
wgtmac, line 46: Why do we need set-scm-tag for `SNAPSHOT` version?

dev/release/release_rc.sh
```diff
@@ -0,0 +1,152 @@
...
+: "${RELEASE_PUSH_TAG:=${RELEASE_DEFAULT}}"
+: "${RELEASE_SIGN:=${RELEASE_DEFAULT}}"
+: "${RELEASE_UPLOAD:=${RELEASE_DEFAULT}}"
+
+cd "${SOURCE_TOP_DIR}"
+
+if [ "${RELEASE_PULL}" -gt 0 ] || [ "${RELEASE_PUSH_TAG}" -gt 0 ]; then
+  git_origin_url="$(git remote get-url origin)"
+  if [ "${git_origin_url}" != "git@github.com:apache/arrow-java.git" ]; then
+    echo "This script must be ran with working copy of apache/arrow-java."
```
wgtmac, line 44: Should we also add that it cannot accept https mode?

dev/release/bump_version.sh
kou: Because we don't create a branch for release in the proposed workflow.

My understanding is that this updates `project.scm.tag`: https://github.com/apache/arrow-java/blob/4a535af9e92b11fc618f15e3dd9cdfad8c554ca2/pom.xml#L84

`pom.xml` in the main branch uses `vX.Y.Z` while we're releasing. We need to back it to `main` after we released `X.Y.Z`.

Do we want to create a release branch? If we have a release branch, we don't need to back it to `main` because `pom.xml` in the `main` branch always uses `main` for `project.scm.tag`. `pom.xml` in release branches use `vX.Y.Z` for `project.scm.tag`.

(If we create a release branch, release process will be a bit complex.)

dev/release/release_rc.sh
kou: Are you (and other commiters/PMC members who may become a release manager) using `https://` for writable clone?
If so, let's accept `https://` too.

FYI: I'm using `https://` for read-only clones.

wgtmac: Usually I use `https://` to git clone a large repo in the beginning which becomes the `origin` remote by default. I can manually switch `origin` to use `ssh` to avoid complication here.

dev/release/bump_version.sh
wgtmac: If we want to follow up with a 19.0.1 release to fix issues in 19.0.0, does it mean that we have to create the RC from main branch too?

dev/release/release_rc.sh
kou: OK. Let's accept `https://` too.
It will not be too complex.

dev/release/bump_version.sh
kou: Yes.

FYI: apache/arrow-julia and apache/arrow-go use this style.

If we have enough resource to maintain multiple major releases (e.g. 18.Y.Z and 19.Y.Z), we can create a release branch for each major release. But I think that we don't have enough resource for now...

wgtmac: Agreed. Let's not over-complicate the script for now.

.github/workflows/rc.yml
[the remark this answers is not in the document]
kou: Are you thinking about a [composite action](https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-composite-action) (e.g. we're using it in apache/arrow: https://github.com/apache/arrow/tree/main/.github/actions/sync-nightlies ) or [reusable workflow](https://docs.github.com/en/actions/sharing-automations/reusing-workflows)?

Anyway, lets' defer this: GH-500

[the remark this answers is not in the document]
lidavidm: I was thinking reusable workflow, but either way works for me - I just mean there's no need to merge/duplicate the pipelines. I'm OK with deferring it, yes.

lidavidm (approved):

kou:
I tried the workflow on my fork. It worked with some fixes.

18.2.0 RC0 GitHub Release: https://github.com/kou/arrow-java/releases/tag/v18.2.0-rc0
18.2.0 GitHub Release: https://github.com/kou/arrow-java/releases/tag/v18.2.0

lidavidm (approved):

dev/release/README.md
```diff
@@ -0,0 +1,147 @@
...
+```
+
+Here is an example to bump version to 19.0.0:
+
+```
+$ GH_TOKEN=${YOUR_GITHUB_TOKEN} dev/release/bump_version.sh 19.0.0
+```
+
+It creates a feature branch and adds a commit that bumps version. This
+opens a pull request from the feature branch by `gh gr create`. So you
```
raulcd, line 48:
```suggestion
opens a pull request from the feature branch by `gh pr create`. So you
```

dev/release/README.md
```diff
@@ -0,0 +1,147 @@
...
+need `gh` command and GitHub personal access token.
+
+See also:
+https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens
+
```
raulcd, line 53: This PR need to be merged before the next step, right? Can we add a comment about it?

dev/release/README.md
```diff
@@ -0,0 +1,147 @@
...
+  * https://dist.apache.org/repos/dist/dev/arrow/KEYS
+  * https://dist.apache.org/repos/dist/release/arrow/KEYS
+
+See the header comment of them how to add a PGP key.
+
+Apache arrow committers can update them by Subversion client with
+their ASF account. e.g.:
+
+```console
+$ svn co https://dist.apache.org/repos/dist/dev/arrow
```
raulcd, line 95: Is release not used?
`https://dist.apache.org/repos/dist/release/arrow/KEYS`?

dev/release/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+Here is an example to release 19.0.0 RC1:
+
+```console
+$ GH_TOKEN=${YOUR_GITHUB_TOKEN} dev/release/release.sh 19.0.0 1
+```
+
+Add the release to ASF's report database via [Apache Committee Report
+Helper](https://reporter.apache.org/addrelease.html?arrow).
+
```
raulcd, line 123: Should we bump again to a SNAPSHOT version once we release?

dev/release/README.md
kou: Right. I should have added a comment about it because it's already pointed out by David... https://github.com/apache/arrow-java/pull/496#discussion_r1909627223
I'll fix this.

dev/release/README.md
kou: Ah, I just realized that I don't know when we should use `dev/arrow/KEYS` and when we should use `release/arrow/KEYS`...

https://infra.apache.org/release-signing.html#keys-policy uses `release/`. So let's use `release/` here.

FYI: I think that PMC members can only change `release/arrow/KEYS` (committers can't change `release/arrow/KEYS`).

dev/release/README.md
kou: Yes! I forgot to add it. I'll add it.

dev/release/README.md
```diff
@@ -0,0 +1,147 @@
...
+opens a pull request from the feature branch by `gh gr create`. So you
+need `gh` command and GitHub personal access token.
+
+See also:
+https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens
+
+### Prepare RC and vote
+
+Run `dev/release/release_rc.sh` on a working copy of
+`git@github.com:apache/arrow-java` not your fork:
```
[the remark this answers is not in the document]
kou: `dev/release/release_rc.sh` checks  `git remote get-url origin`. So if we use a wrong clone, the script stops with an error message.

dev/release/README.md
kou: Ah, this command lines are just for example how to update `KEYS`. (See the above "e.g.:" part.) So the current description doesn't say that we should update only `dev/arrow/KEYS`.

kou: But `editor KEYS` isn't a good example. I'll fix it.

raulcd (approved):
:+1:

PR #501, opened 2025-01-10
MINOR: Bump version to 19.0.0

kou:


kou:
We will not release a new version immediately but can we merge this to main for easy to GH-496 test?

lidavidm (approved):
I'm OK merging this to check CI. Is the plan to revert it afterwards?

We may want to start another discussion about this but I wonder if we should start following SemVer here now that we're separate from Arrow C++. (So the next actual release would be 18.2.0.)

kou:
Ah, OK. I'll use 18.2.0 instead of this.
If we use 18.2.0, we don't need to revert this, right?

kou:
GH-502

lidavidm:
Oh, I would've also been OK with 19.0.0 -> 19.1.0 or something. Either way I'll approve the other PR.

PR #502, opened 2025-01-10
MINOR: Bump version to 18.2.0

kou:


lidavidm (approved):

lidavidm:
I asked about reverting since, if we do actually run the RC release job nightly, and it uploads to a GitHub release, we either need to put `-SNAPSHOT` back in the version or we need to make sure the GitHub release is clearly marked as a draft/prerelease version.

kou:
> if we do actually run the RC release job nightly, and it uploads to a GitHub release, we either need to put `-SNAPSHOT` back in the version or we need to make sure the GitHub release is clearly marked as a draft/prerelease version.

Ah, I understand your concern!
The nightly RC workflow doesn't uploads artifacts to a GitHub release.
Because the upload job is used only for tag push: https://github.com/apache/arrow-java/pull/496/files#diff-2ff8a59693d8d05dd67560f01c2a49d9e05f4b1ccefeb91f8486744532214e44R113

kou:
Anyway, I merge this for testing.

Issue #445, opened 2024-12-03
[Java][Flight] Add middleware for auth2 bearer tokens


devinrsmith:
### Describe the enhancement requested

This would be similar in spirit to `org.apache.arrow.flight.client.ClientCookieMiddleware`, essentially echo-ing back the latest bearer token for authentication purposes. This would also allow for authentication schemes where the server rotates the bearer token.

### Component(s)

Java

PR #503, opened 2025-01-10
GH-445: [Flight] Add middleware for auth2 bearer tokens

devinrsmith:
### Rationale for this change

This makes it easier to work with auth2 from the Java client without needing to resort to ClientCookieMiddleware (ie, don't need to force the server to support a non-standard auth scheme). I think it could be highly beneficial for ease-of-use when directly creating a FlightClient, and I would expect that the various FlightSqlClient drivers (ADBC, JDBC) could benefit from supporting it as a configuration option.

A new FlightClientMiddleware had been added.

### Are these changes tested?

These changes have been manually tested; if more testing is necessary, I'm hoping another contributor can provide it or guide me through a test plan.

Note: this is a re-creation of https://github.com/apache/arrow-java/pull/446, I think some codeowner logic was not in place when I originally created the PR.

### DIFF
--- a/flight/flight-core/src/main/java/org/apache/arrow/flight/auth2/ClientBearerMiddleware.java
+++ b/flight/flight-core/src/main/java/org/apache/arrow/flight/auth2/ClientBearerMiddleware.java
@@ -1,0 +1,77 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.arrow.flight.auth2;
+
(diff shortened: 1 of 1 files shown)


flight/flight-core/src/main/java/org/apache/arrow/flight/auth2/ClientBearerMiddleware.java
```diff
@@ -0,0 +1,77 @@
...
+          Auth2Constants.AUTHORIZATION_HEADER, Auth2Constants.BEARER_PREFIX + bearerValue);
+    }
+  }
+
+  @Override
+  public void onCallCompleted(CallStatus status) {}
+
+  public static class Factory implements FlightClientMiddleware.Factory {
+
+    private volatile String bearerValue;
```
lidavidm, line 57: I think AtomicReference would be more appropriate here

flight/flight-core/src/main/java/org/apache/arrow/flight/auth2/ClientBearerMiddleware.java
```diff
@@ -0,0 +1,77 @@
...
+    public FlightClientMiddleware onCallStarted(CallInfo info) {
+      return new ClientBearerMiddleware(this);
+    }
+
+    String getBearerValue() {
+      return bearerValue;
+    }
+
+    void setBearerValue(String bearerValue) {
+      // Avoid volatile write if the value hasn't changed
```
lidavidm, line 71: Doesn't this optimization introduce a TOC/TOU error?

lidavidm (commented):
I think this might be useful, but we probably need to sit down and sort out all the existing auth options...there are too many at this point and it's unclear how each of them work without having to go read every implementation (I _think_ this PR does not overlap with any of them, but it's hard to make sure)

I think this needs at least a basic test using an in-process server. For instance the server could track the last token it received, and the test could assert that the client and the server see the 'correct' token as it is changed (on either side).

PR #517, opened 2025-01-15
GH-523: [Release] Add support for building .jar including JNI binaries

kou:
Fixes GH-523.

Merge `test_jni.yml` to `rc.yml` so that .jar including JNI binaries can be built and uploaded to GitHub Releases for RC.

ccache support is also enabled for fast CI.


.github/workflows/rc.yml
```diff
@@ -70,15 +70,269 @@ jobs:
...
+      - name: Download source archive
+        uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16 # v4.1.8
+        with:
+          name: release-source
+      - name: Extract Download the latest Apache Arrow C++
+        run: |
+          tar -xf apache-arrow-java-*.tar.gz --strip-components=1
+      - name: Download the latest Apache Arrow C++
+        run: |
+          ci/scripts/download_cpp.sh
```
kou, line 108: We may want to use the main branch instead of related Apache Arrow C++ on daily run. But it's out-of-scope in this PR.

.github/workflows/rc.yml
```diff
@@ -70,15 +70,269 @@ jobs:
...
+      MACOSX_DEPLOYMENT_TARGET: "14.0"
+    steps:
+      - name: Download source archive
+        uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16 # v4.1.8
+        with:
+          name: release-source
+      - name: Extract Download the latest Apache Arrow C++
+        run: |
+          tar -xf apache-arrow-java-*.tar.gz --strip-components=1
+      # We need 19.0.0 for latest Boost support
```
kou, line 165: apache/arrow#45053

.github/workflows/rc.yml
```diff
@@ -105,37 +357,40 @@ jobs:
...
+          export VERIFY_BINARY=1
+          export VERIFY_SOURCE=1
+          cd apache-arrow-java-*
+          mv ../${tar_gz} ./
+          mv ../release-binaries binaries
+          dev/release/verify_rc.sh "${version}" "${rc}"
   upload:
     name: Upload
-    if: github.ref_type == 'tag'
+    # if: github.ref_type == 'tag'
```
kou, line 370: I'll revert this after I confirmed that this job works.

.github/workflows/rc.yml
```diff
@@ -105,37 +357,40 @@ jobs:
...
+          pattern: release-*
+          path: artifacts
       - name: Upload
         run: |
           # GH-499: How to create release notes?
           version=${GITHUB_REF_NAME%-rc*}
           version=${version#v}
           rc=${GITHUB_REF_NAME#*-rc}
-          gh release create ${GITHUB_REF_NAME} \
+          echo gh release create ${GITHUB_REF_NAME} \
```
kou, line 388: I'll revert this after I confirmed that this job works.

.github/workflows/rc.yml
```diff
@@ -70,15 +70,269 @@ jobs:
...
+      - name: Checkout apache/arrow-testing
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          repository: apache/arrow-testing
+          path: arrow/testing
+      - name: Checkout apache/parquet-testing
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          repository: apache/parquet-testing
+          path: arrow/cpp/submodules/parquet-testing
```
lidavidm, line 118: Do we need testing data? (I don't think we need to run C++ tests here?)

lidavidm (approved):

.github/workflows/rc.yml
kou: OK. I'll disable C++ tests.
(I thought that we need to run C++ tests because it seems that we don't have enough test for JNI related codes.)

lidavidm: The C++ tests we run here are just the same tests that apache/arrow already runs right? (I agree we need more testing for the JNI code itself, though...)

kou: Right but we use different build options that aren't covered by apache/arrow.

lidavidm: Hmm if they're that different then should we be testing the config upstream? Or otherwise in that case I don't mind running the C++ tests here if our config is very different (it mostly just adds a lot of build time)

kou: OK. Let's do it in upstream.

.github/workflows/rc.yml
kou: Reverted. This worked: https://github.com/apache/arrow-java/actions/runs/12781206977/job/35629480932#step:3:17

kou:
CI passed. This is ready.

kou:
I'll merge this today.

I'll work on the followings as separated tasks:
* Sign uploaded `.jar` and related files
* Upload signed `.jar` and related files to staging
* Publish files on staging after release vote passed

Issue #518, opened 2025-01-15
[CI] Cleanup apache-arrow-java.tar.gz created by pre-commit


kou:
### Describe the enhancement requested

`pre-commit run` creates `apache-arrow-java.tar.gz` but it's not removed. So `apache-arrow-java.tar.gz` exists after `pre-commit run`.

PR #519, opened 2025-01-15
GH-518: Cleanup apache-arrow-java.tar.gz created by pre-commit

kou:
Fixes GH-518.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -35,7 +35,8 @@ repos:
             git archive HEAD \
               --prefix=apache-arrow-java/ \
               --output=apache-arrow-java.tar.gz && \
-              dev/release/run_rat.sh apache-arrow-java.tar.gz"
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -35,7 +35,8 @@ repos:
             git archive HEAD \
               --prefix=apache-arrow-java/ \
               --output=apache-arrow-java.tar.gz && \
-              dev/release/run_rat.sh apache-arrow-java.tar.gz"
+              dev/release/run_rat.sh apache-arrow-java.tar.gz && \
+              rm -f apache-arrow-java.tar.gz"
```
lidavidm, line 39: I guess it will get left behind if run_rat fails, but I suppose that's ok?

lidavidm (approved):

.pre-commit-config.yaml
kou: Yes. In the case, we should check `apache-arrow-java.tar.gz`.

Issue #527, opened 2025-01-16
[Release] Add support for .jar for Windows


kou:
### Describe the enhancement requested

https://github.com/apache/arrow/blob/6a0414bd9a91e890ec6a45369bf61f405180628c/dev/tasks/java-jars/github.yml#L161-L193

PR #528, opened 2025-01-17
GH-527: [Release] Add support for .jar for Windows

kou:
Fixes GH-527.

It seems that arrow-testing with ARROW_BUILD_STATIC=ON and GTest_SOURCE=BUNDLED is broken. We'll enable JNI test after we fix it in apache/arrow.

### DIFF
--- a/.github/workflows/rc.yml
+++ b/.github/workflows/rc.yml
@@ -76,7 +76,7 @@ jobs:
           name: release-source
           path: |
             apache-arrow-java-*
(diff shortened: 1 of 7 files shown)


kou:
CI is green.

wgtmac (approved):
+1

I just did a quick scan (but may not be able to spot any issue).

lidavidm (approved):
Thanks!

Issue #537, opened 2025-01-20
[Release][Docs] Generate API reference


kou:
### Describe the enhancement requested

We can generate API reference by javadoc.

kou:
Done by GH-538.

Issue #67, opened 2024-07-11
[Java] Do not publish modules based on filenames


danepitkin:
### Describe the bug, including details regarding any error messages, version, and platform.

Arrow Java should not publish modules based on filenames. Flatbuffers/Protobuf can probably be shaded to avoid this. We can potentially remove usage of jsr305 in favor of another library.

```
[WARNING] * Required filename-based automodules detected: [flatbuffers-java-24.3.25.jar, jsr305-3.0.2.jar, protobuf-java-3.25.1.jar, protobuf-java-util-3.25.1.jar]. Please don't publish this project to a public artifact repository! *
```

- [x] https://github.com/apache/arrow/issues/43396
- [x] https://github.com/apache/arrow/issues/43437
- [ ] https://github.com/google/flatbuffers/issues/8348

### Component(s)

Java

PR #538, opened 2025-01-20
GH-537: [Release][Docs] Generate API reference

kou:
The followings are out-of-scope:

* Generate Sphinx based documents
* Publish this to https://arrow.apache.org/java/reference/ in release process
* Publish this to https://arrow.apache.org/dev/reference/ nightly

### DIFF
--- a/.github/workflows/rc.yml
+++ b/.github/workflows/rc.yml
@@ -76,6 +76,49 @@ jobs:
           name: release-source
           path: |
             apache-arrow-java-*
+  docs:
+    name: Docs
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    needs:
+      - source
+    env:
+      DOCKER_VOLUME_PREFIX: .docker/
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - name: Download source archive
+        uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16 # v4.1.8
+        with:
+          name: release-source
+      - name: Extract source archive
+        run: |
+          tar -xf apache-arrow-java-*.tar.gz --strip-components=1
+      - uses: docker/login-action@9780b0c442fbb1117ed29e0efdff1e18412f7567 # v3.3.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Cache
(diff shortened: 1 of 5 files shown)


.github/workflows/rc.yml
lidavidm, line 1: Does this build the docs for the JNI components too? Or do we need to enable the JNI build to get the Javadocs?

ci/scripts/build.sh
```diff
@@ -61,27 +59,38 @@ done
 
 pushd "${build_dir}"
 
-if [ "${ARROW_JAVA_SHADE_FLATBUFFERS}" == "ON" ]; then
+# TODO: ARROW_JAVA_SHADE_FLATBUFFERS isn't used for our artifacts. Do
+# we need this?
+# See also: https://github.com/apache/arrow/issues/22021
+if [ "${ARROW_JAVA_SHADE_FLATBUFFERS:-OFF}" == "ON" ]; then
```
lidavidm, line 65: I think we need to revisit this sometime because of https://github.com/apache/arrow-java/issues/67 (and it appears Google has closed the issue asking them to fix it upstream). 

AFAIK the discussion in the issue you found is weird...We shouldn't need to change imports ourselves, the shading should take care of that. But that's not relevant here.

lidavidm (approved):

.github/workflows/rc.yml
kou: > Does this build the docs for the JNI components too?

No.

> Or do we need to enable the JNI build to get the Javadocs?

Yes. But it's easy because we already have JNI enabled jobs. I'll do it.

ci/scripts/build.sh
kou: OK. Can we use #67 for this topic? Or should we create a separated issue for this topic?

lidavidm: Yes, we can use #67!

lidavidm (approved):

.github/workflows/rc.yml
kou: I thought we can use `docker compose run vcpkg-jni` but it doesn't create JAR. So I can't use it for `mvn site`.
I used `ci/scripts/jni_full_build.sh` instead.

kou:
I confirmed that `release-docs` includes API reference for JNI such as C and Gandiva: https://github.com/apache/arrow-java/actions/runs/12863727769?pr=538#artifacts

lidavidm (approved):

PR #557, opened 2025-01-23
GH-41: BaseVariableWidthViewVector setZero only if necessary

### LINKED ISSUE #41: [Java] VariableWidthViewVectorBenchmarks
The performance of  VariableWidthViewVector is much slower than VariableWidthVector in some cases. `setZero` for overwriting is the bottleneck.
```
VariableWidthViewVectorBenchmarks.setSafeFromArray       1  avgt    5  0.088 ± 0.002  ms/op
VariableWidthViewVectorBenchmarks.setSafeFromArray       2  avgt    5  0.095 ± 0.020  ms/op
VariableWidthViewVectorBenchmarks.setSafeFromArray      10  avgt    5  0.091 ± 0.012  ms/op
VariableWidthViewVectorBenchmarks.setSafeFromArray      40  avgt    5  0.087 ± 0.001  ms/op
    VariableWidthVectorBenchmarks.setSafeFromArray       1  avgt    5  0.010 ±  0.001  ms/op
    VariableWidthVectorBenchmarks.setSafeFromArray       2  avgt    5  0.013 ±  0.001  ms/op
    VariableWidthVectorBenchmarks.setSafeFromArray      10  avgt    5  0.032 ±  0.001  ms/op
    VariableWidthVectorBenchmarks.setSafeFromArray      40  avgt    5  0.105 ±  0.002  ms/op
```
image

Environment: 2.6 GHz Intel Core i7, MacOS 13.6, openjdk version "13.0.2"
See code of benchmark in https://github.com/apache/arrow/pull/44634. 

### Component(s)

Java

ViggoC:
Closes #41.

### DIFF
--- a/vector/src/main/java/org/apache/arrow/vector/BaseVariableWidthViewVector.java
+++ b/vector/src/main/java/org/apache/arrow/vector/BaseVariableWidthViewVector.java
@@ -1367,11 +1367,13 @@ public abstract class BaseVariableWidthViewVector extends BaseValueVector
   protected final void setBytes(int index, byte[] value, int start, int length) {
     int writePosition = index * ELEMENT_SIZE;
 
-    // to clear the memory segment of view being written to
-    // this is helpful in case of overwriting the value
-    viewBuffer.setZero(writePosition, ELEMENT_SIZE);
-
     if (length <= INLINE_SIZE) {
+      // to clear the memory segment of view being written to
+      // if it has been set
+      if (viewBuffer.getLong(writePosition) != 0 || viewBuffer.getLong(writePosition + 8) != 0) {
+        viewBuffer.setZero(writePosition, ELEMENT_SIZE);
+      }
+
       // allocate inline buffer
       // set length
       viewBuffer.setInt(writePosition, length);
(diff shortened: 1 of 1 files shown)


vector/src/main/java/org/apache/arrow/vector/BaseVariableWidthViewVector.java
```diff
@@ -1367,11 +1367,13 @@ protected ArrowBuf allocateOrGetLastDataBuffer(int length) {
...
     int writePosition = index * ELEMENT_SIZE;
 
-    // to clear the memory segment of view being written to
-    // this is helpful in case of overwriting the value
-    viewBuffer.setZero(writePosition, ELEMENT_SIZE);
-
     if (length <= INLINE_SIZE) {
+      // to clear the memory segment of view being written to
+      // if it has been set
+      if (viewBuffer.getLong(writePosition) != 0 || viewBuffer.getLong(writePosition + 8) != 0) {
```
lidavidm, line 1373: So the basic effect is that if the view buffer is all 0, we skip setting it back to 0? And basically, for freshly allocated memory, this skips a redundant memset? Can we explain it a little more in the comment?

ViggoC: Yes, it's redundant to setZero if the data is already zero. Checking will introduce some overhead, but it is much smaller than that of setMemory.

ViggoC: Update in 16c9e2f4bc5921a4d6e0572954b19b5f958dbf85

wgtmac: Is it safe to check only these positions? IIUC, ZERO is a valid value to be set.

lidavidm: @wgtmac are we on the same page here?

wgtmac: Yes, it sounds good.

Issue #548, opened 2025-01-21
[Release][Docs] Publish docs to https://arrow.apache.org/java/current/ on release


kou:
### Describe the enhancement requested

ADBC uses https://arrow.apache.org/adbc/{current,${VERSION}} such as:
* https://arrow.apache.org/adbc/current/
* https://arrow.apache.org/adbc/15/

We can use the same approach.

ADBC does this in https://github.com/apache/arrow-adbc/blob/922a0f2b3f90267d2dbe7ff89a09d8dc9cf6d1a5/ci/scripts/website_build.sh#L49-L64 .

PR #569, opened 2025-01-28
GH-548: [Release][Docs] Publish docs to https://arrow.apache.org/java/current/ on release

kou:
Fixes GH-548.

### DIFF
--- a/.github/workflows/rc.yml
+++ b/.github/workflows/rc.yml
@@ -536,7 +536,7 @@ jobs:
           cd site
           git config --global user.name 'github-actions[bot]'
           git config --global user.email 'github-actions[bot]@users.noreply.github.com'
-          git commit -m "Publish documents (${GITHUB_SHA})"
+          git commit -m "Publish documentation (${GITHUB_SHA})"
           git push origin "${BRANCH}"
   verify:
     name: Verify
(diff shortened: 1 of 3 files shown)


.github/workflows/rc.yml
```diff
@@ -537,7 +536,7 @@ jobs:
           cd site
           git config --global user.name 'github-actions[bot]'
           git config --global user.email 'github-actions[bot]@users.noreply.github.com'
-          git commit -m "Publish documentation (${GITHUB_SHA})"
```
lidavidm, line 540: nit, but "documentation" is perfectly fine here!

lidavidm (approved):

.github/workflows/rc.yml
kou: Oh! I'll revert this change!

(I'm not sure the difference in "docs", "document", "documents", "documentation", "documentations", ...)

lidavidm: docs -> short for documentation or documents (confusingly...)
document -> 書類など
documents -> plural of document (but usually not used for documentation, even though docs is used as the short form of both...)
documentations -> I think this not used (though apparently not _wrong_, just very rare)

kou: Thanks!!!

lidavidm (approved):

kou:
Let's try this in a real release process.

PR #578, opened 2025-01-31
GH-579: Add and fix LICENSE.txt and NOTICE.txt in the distributed artifacts

### LINKED ISSUE #579: LICENSE.txt/NOTICE.txt miss dependencies for binary artifacts

The current LICENSE.txt/NOTICE.txt are valid only for tar.gz.
jars redistribute dependencies. So we need to mention them.
See also: https://infra.apache.org/licensing-howto.html#binary

jbonofre:
Fixes #579.

In our distributed artifacts, especially the shading ones, we have to provide `LICENSE.txt` and `NOTICE.txt` with all "bundled" dependencies.

@lidavidm @kou This is a draft PR as I have to fix the non-shading artifacts.


jbonofre:
@kou @lidavidm if you have time, you can already take a look on the content. Please let me know if you have any question.

lidavidm (approved):

lidavidm:
BTW, how were these generated? Or do they have to be assembled by hand?

jbonofre:
> BTW, how were these generated? Or do they have to be assembled by hand?

For this release, I did that "manually". After the release, I will add the maven tooling to at least check the content.

kou (approved):
+1

Thanks!

LICENSE.txt
```diff
@@ -202,16 +202,21 @@
    limitations under the License.
 
 --------------------------------------------------------------------------------
-vector/src/main/java/org/apache/arrow/vector/util/IntObjectHashMap.java
-vector/src/main/java/org/apache/arrow/vector/util/IntObjectMap.java
+This product includes code from Netty 4.1.117.Final:
```
wgtmac, line 205: Is it required to include the specific version? It would be easily out of sync since the dependabot does not maintain this.

jbonofre: Unfortunately (or not 😄 ), yes,  It's important to specify the version of the dependency as licenses sometimes change as product versions change.
Also the `NOTICE` can change from a version to another.
So, it's better to document to actual version bundled in our distributed jar.

So source distribution, we should document the version where the code has been copied from.

jbonofre:
I'm fixing the location and other jar resources.

kou:
Should we merge PRs from Dependabot such as https://github.com/apache/arrow-java/pull/584 after the next release?
I think that they may change dependency versions in binary artifacts.

jbonofre:
> Should we merge PRs from Dependabot such as #584 after the next release? I think that they may change dependency versions in binary artifacts.

We can merge dependabot PRs, I will update this PR accordingly. That's totally fine for me. I need ~ 1 day to fix the last "jar" creation.

kou:
OK. I'll merge the following PRs that were opened in this week:

* #580 
* #581 
* #582 
* #583 
* #584

Our Dependabot configuration uses "weekly". So the next PRs from Dependabot will be opened in the next week.

jbonofre:
@kou thanks ! I'm updating this PR accordingly.

jbonofre:
I'm fixing the endline thing.

LICENSE.txt
wgtmac: Thanks for the explanation! If we don't have good automation tools to keep them in sync, at least we need to make sure they are accurate in the release process.

lidavidm (approved):

kou:
@jbonofre Can we merge this?

jbonofre:
> @jbonofre Can we merge this?

Yes, we are good 👍

jbonofre:
@lidavidm @kou thanks for the review and merge. I'm doing a quick new run but I think we are good for the release.

kou:
Thanks! I'll create cut an RC.

Issue #588, opened 2025-02-05
[Release] Increase wait interval in `dev/release/release_rc.sh`


kou:
### Describe the enhancement requested

`release.yml` takes at least 30min: https://github.com/apache/arrow-java/actions/runs/13151205894/usage

So the current 60s wait interval is too short. We can use 300s or 600s.

PR #589, opened 2025-02-05
GH-588: [Release] Increase wait interval in `dev/release/release_rc.sh`

kou:
Fixes GH-588.

`release.yml` takes at least 30min now. So 60s wait interval is too short. We can use longer interval to avoid needless API requests.

### DIFF
--- a/dev/release/release_rc.sh
+++ b/dev/release/release_rc.sh
@@ -109,13 +109,13 @@ if [ "${RELEASE_SIGN}" -gt 0 ]; then
     if [ -n "${run_id}" ]; then
       break
     fi
-    sleep 60
+    sleep 600
   done
 
   echo "Found GitHub Actions workflow with ID: ${run_id}"
   gh run watch \
     --exit-status "${run_id}" \
-    --interval 60 \
+    --interval 600 \
     --repo "${repository}"
 
   echo "Downloading artifacts from GitHub Releases"


jbonofre (approved):

dev/release/release_rc.sh
```diff
@@ -109,13 +109,13 @@ if [ "${RELEASE_SIGN}" -gt 0 ]; then
     if [ -n "${run_id}" ]; then
       break
     fi
-    sleep 60
+    sleep 600
```
lidavidm, line 112: We have to wait 10 minutes!?

lidavidm (approved):

dev/release/release_rc.sh
jbonofre: As it's a `sleep`, it seems so.  It's maybe related to @kou network. 10mn is OK though is necessary.

lidavidm: Ah the issue explains the whole process takes ~30 minutes so ~10 minutes seems reasonable (I guess Github doesn't like pinging too often)

jbonofre: Yes, it's pretty clear regarding https://github.com/apache/arrow-java/issues/588#issue-2832065440

kou: Yes.
I've added it to this PR description too. Sorry. I should have explained it in the PR description too.

PR #590, opened 2025-02-06
MINOR: Upload release artifacts individually with sleep

lidavidm:
According to https://github.com/cli/cli/issues/9586, GitHub CLI doesn't respect the GitHub rate limits and GitHub is uninterested in fixing this.


kou (approved):
+1

Good catch! Let's try this!

Issue #601, opened 2025-02-12
[Java][Gandiva] Synchronize some methods on Projector.


lriggs:
### Describe the enhancement requested

Our product, which uses Gandiva, was encountering errors from LLVM during intense query processing stress tests. I traced the issue back to the JIT upgrade (https://github.com/apache/arrow/pull/39098), but it seems it only started occurring then because the new JIT engine requires unique symbol names unlike the old engine.

The error from llvm is:   Duplicate definition of symbol 'expr_0_0'.

After updating the llvm version to 18 (https://github.com/apache/arrow/pull/45114) you can see this error occasionally when running the Gandiva ProjectorTest.

```
Failed to make LLVM module due to Failed to add IR module to LLJIT: Duplicate definition of symbol 'expr_0_0'

	at org.apache.arrow.gandiva.evaluator.JniWrapper.buildProjector(Native Method)
	at org.apache.arrow.gandiva.evaluator.Projector.make(Projector.java:206)
	at org.apache.arrow.gandiva.evaluator.Projector.make(Projector.java:70)
	at org.apache.arrow.gandiva.evaluator.ProjectorTest.lambda$testMakeProjectorParallel$1(ProjectorTest.java:138)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)
[WARNING] Tests run: 44, Failures: 0, Errors: 0, Skipped: 1, Time elapsed: 6.656 s -- in org.apache.arrow.gandiva.evaluator.ProjectorTest
```

Adding synchronization the the make and evaluate methods in gandiva/src/main/java/org/apache/arrow/gandiva/evaluator/Projector.java will fix the issue.

lriggs:
```
diff --git a/java/gandiva/src/main/java/org/apache/arrow/gandiva/evaluator/Projector.java b/java/gandiva/src/main/java/org/apache/arrow/gandiva/evaluator/Projector.java
index 951cebc88e..a67d20db6b 100644
--- a/java/gandiva/src/main/java/org/apache/arrow/gandiva/evaluator/Projector.java
[...]

PR #595, opened 2025-02-08
GH-55: [Gandiva] Re-enable tests

lriggs:
Fixes GH-55.

### Rationale for this change
I fixed the linking error here: https://github.com/apache/arrow/pull/45114/

### What changes are included in this PR?
Reenabling some disabled tests.

### Are these changes tested?
Yes, they are tests.

### Are there any user-facing changes?
No.

PR #602, opened 2025-02-12
GH-601: [Gandiva] Synchronize some methods on the Projector

lriggs:
### Rationale for this change
Multiple threads can attempt to create the same llvm expression in Gandiva. This isn't allowed with the new JIT compiler, so synchronizing will prevent this scenario.

Synchronize some methods to avoid adding duplicate llvm expressions.

### Are these changes tested?
Yes, through unit tests in Gandiva.

No.

Closes GH-601

### DIFF
--- a/gandiva/src/main/java/org/apache/arrow/gandiva/evaluator/Projector.java
+++ b/gandiva/src/main/java/org/apache/arrow/gandiva/evaluator/Projector.java
@@ -188,7 +188,7 @@ public class Projector {
    * @param configurationId Custom configuration created through config builder.
    * @return A native evaluator object that can be used to invoke these projections on a RecordBatch
    */
-  public static Projector make(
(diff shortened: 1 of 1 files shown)


lidavidm:
@lriggs since you've been working on Gandiva lately, would you be able to help take a look at the changes here?

kou:
We should rebase on main after we merge GH-595.

lidavidm:
> @lriggs since you've been working on Gandiva lately, would you be able to help take a look at the changes here?

Ok I'm clearly not paying attention...you are lriggs :grimacing: 

As Kou states, let's rebase this after the other PR.

kou:
Could you rebase on main?

lriggs:
I rebased but don't seem to have permissions to add the bug-fix label.

lidavidm (approved):

Issue #611, opened 2025-02-13
Documentation uses "18.1.0" not "18.2.0"


kou:
### Describe the bug, including details regarding any error messages, version, and platform.

https://arrow.apache.org/java/18.2.0/ shows "arrow-java 18.1.0 documentation".

kou:
We need to update https://github.com/apache/arrow-java/blob/34e2b08ff1b06c864f63f3ad1a3a771d0e7b058b/docs/source/conf.py#L29 by `dev/release/bump-version.sh`.

PR #612, opened 2025-02-13
GH-611: Update docs version in release process

kou:
Fixes GH-611.


lidavidm (approved):

lidavidm:
In theory...we could change conf.py to read it out of the pom.xml...

kou:
Oh, it's a good idea! Let's use the approach.

kou:
I close this in favor of #613.

PR #613, opened 2025-02-13
GH-611: [Docs] Detect version from pom.xml

kou:
Fixes GH-611.


lidavidm (commented):
Python does have built-in XML processing, so we can avoid regexps:

```python
import xml.etree.ElementTree as ET
tree = ET.parse("pom.xml")
ns = {"maven": "http://maven.apache.org/POM/4.0.0"}
version = tree.getroot().find("maven:version", ns).text.strip()
print(version)
```

kou:
Thanks! I didn't know it!

lidavidm (approved):

PR #627, opened 2025-02-21
GH-625: Map MinorType getNewFieldWriter returns UnionMapWriter

### LINKED ISSUE #625: getNewFieldWriter returns UnionListWriter instead of UnionMapWriter
Not sure if I am using this wrong, but I noticed `getNewFieldWriter` for the minorType `MAP` is returning a `UnionListWriter` instead of a `UnionMapWriter`.

https://github.com/apache/arrow-java/blob/799d9fd45e36d962c7a7bae0f3903bb414062554/vector/src/main/java/org/apache/arrow/vector/types/Types.java#L722-L726

This leads to the issue of this code block throwing an error:
```java
MinorType.MAP.getNewFieldWriter(myMapVector)
writer.startMap()
```
Exception:
```
You tried to start a map key when you are using a ValueWriter of type UnionListWriter.
java.lang.IllegalStateException: You tried to start a map key when you are using a ValueWriter of type UnionListWriter.
	at org.apache.arrow.vector.complex.impl.AbstractFieldWriter.key(AbstractFieldWriter.java:123)
	at org.apache.arrow.vector.complex.impl.UnionListWriter.key(UnionListWriter.java:69)
...
```

I believe this code can be easily fixed by just changing it to:
```java
   @Override 
   public FieldWriter getNewFieldWriter(ValueVector vector) { 
     return new UnionMapWriter((MapVector) vector); 
   } 
```

wsuppiger:
## What's Changed

Closes #625

### DIFF
--- a/vector/src/main/java/org/apache/arrow/vector/types/Types.java
+++ b/vector/src/main/java/org/apache/arrow/vector/types/Types.java
@@ -116,6 +116,7 @@ import org.apache.arrow.vector.complex.impl.UInt8WriterImpl;
 import org.apache.arrow.vector.complex.impl.UnionLargeListViewWriter;
 import org.apache.arrow.vector.complex.impl.UnionLargeListWriter;
 import org.apache.arrow.vector.complex.impl.UnionListWriter;
+import org.apache.arrow.vector.complex.impl.UnionMapWriter;
 import org.apache.arrow.vector.complex.impl.UnionWriter;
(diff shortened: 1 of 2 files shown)


lidavidm (commented):
Thanks! Would it be possible to add a unit test somewhere?

lidavidm (approved):
Thanks!

Issue #108, opened 2024-04-02
[Java] Type-ids in UnionVector are erroneously coupled to the Arrow types of the underlying vectors


jarohen:
### Describe the bug, including details regarding any error messages, version, and platform.

re: https://lists.apache.org/thread/z89xlvw7v1rwq89gknflhsj3c65x20kd

It seems that the UnionVector implementation (particularly `initializeChildrenFromFields` (#29848), `getVector`, `getVectorByType`, `setSafe` etc) assumes that the type-id is always based on the ArrowType, but the Schema.fbs spec is more lenient - users have the choice to use whatever type-ids they require.

For example, in XTDB, we're trying to represent an algebraic data type (ADT) of 'put', 'delete' and 'erase' events as a sparse union. Delete and erase have the same type, so UnionVector currently expects them to be the same type-id (whereas, in DenseUnionVector, we can use type-ids 0, 1 and 2).

Would there be an appetite for (potentially relatively significant) changes to UnionVector to make it behave this way? We could perhaps consider bringing it more in line with DenseUnionVector which seems closer to the spec. Would be happy to work on it if so.

Cheers,

James/Finn (@FiV0)

### Component(s)

Java

nbauernfeind:
Are there any plans on fixing this? It's inconvenient that `UnionVector#setType` and `DenseUnionVector#setTypeId` deviate from one another.

lidavidm:
Unfortunately there's just very little maintainer power for arrow-java and what's left is mostly focused on trying to keep things running (e.g. trying to set up release processes again now that we've split the repo) :/

lidavidm:
It frustrates me too and I wish I had time to help out and fix things, improve the library, keep Arrow abreast of things like the new native memory interface, etc. but that's just not happening with my current development time/goals

nbauernfeind:
Thanks for sharing; I totally understand the very real issue of not having enough resources.

lidavidm:
It's doubly frustrating because I feel like the Java library gets certain design decisions right vs the other implementations 😬 

Unfortunately we probably need some sponsor or champion to fund development...The lack of maintainers is also sort of self-perpetuating since there's not many people to review things in the first place

jbonofre:
Maybe I can take a look after the releases. I think arrow-Java needs some love (and I have a lot of love to give :)).

martin-traverse:
[...]

PR #638, opened 2025-02-26
GH-615: Produce Avro core data types out of Arrow VSR

### LINKED ISSUE #615: Avro write support - will you accept a patch?

Hi,

We need full Avro read / write support in our project and are working on an implementation. I had a look at what already exists in arrow-java, I think it would be fairly straightforward to extend what is there to get full read/write support in the Arrow Java project. Here is what I am proposing:

* A set of producers to handle the Avro data structures, mirroring the existing consumers
* Handle the high level file structure (header, embedded schema and block structure)
* Support for compressed blocks (using the existing codecs in the Avro project)
* High level APIs for read / write, including incremental read (block by block, corresponding to the VSR)

The last point is important for us because we handle streaming data, if we can check a whole block is available before reading it we should be able to prevent avoid on IO calls.

If I draft a PR along these lines, would there be interest to help me refine it and get it into arrow-java? If not we can do our own implementation which will be simpler because we don't need all the features and data types, but I think the delta is not that large and IMO it would be a good thing to have in the Arrow Java toolkit.

Thoughts welcome!

martin-traverse:
## What's Changed

Per discussion in #615 , here is a first take on the core producers to generate Avro data from Arrow vectors. There are a few points I'd like to clarify before going further:

* Nullability. Avro only understands nullable types as unions but that is not normally how they will be if the data comes from other sources. I have added a special NullableProducer to handle nullable vectors which are not unions. We will need something equivalent in the consumers and probably a setting in the AvroToArrowConfig to control it on read, defaulting to current behaviour. I have also added special handling for nullable unions, because unions can't be nested  (i.e. you can't nest "type | null" as a type inside a union). I can add consumers to handle both (unions and regular types) for review, if that sounds right? At the moment the schema for nullable fields gets quite mangled on a round trip!

* Arrow has a lot more types than Avro, at the level of minor / vector types. Going Avro -> Arrow we just pick the direct equivalent. Going Arrow -> Avro, we could cast silently if there is no loss of precision. E.g. TinyInt and SmallInt -> Int and so on. For types like e.g. Decimal256 and LargeVarChar we could write out safely but would need support in the consumers to read back the wider types. I could start by adding the safe conversions now and we could come back to the wide types in a later PR maybe?

* I have made Producer inherit AutoClosable the same as Consumer does. Not sure if that is always what you want though - it will free the buffers in the VSR, but you might want to keep the VSR after you are finished with IO. Do we need something like detachValueVector() to go with resetValueVector()? Calling close() after detach would not affect the vector.

* Type information is inferred from the list of vectors, using minor types. We'll also need to generate the Avro schema, the input for that would be a list of fields. I haven't done it yet but will do if that sounds right.

* Dictionary encoding for enums not implemented yet, I'll add it if the rest looks good. Caveat is that dictionaries must be fixed before the encoding starts if we are writing out the whole file in one go (i.e. if the Avro schema is at the start of the container file). If the schema is saved separately that limitation need not apply, we could provide the schema once encoding is finished.

Please do let me know if this is going in the right direction + any comments. If it is I will add the missing pieces and start the exhaustive test coverage to mirror the consumers. Once it's done this PR should get us to the point where we can round trip the contents of an individual block for most data types, but it does not address the container format.

Closes #615.


adapter/avro/src/main/java/org/apache/arrow/adapter/avro/producers/Producer.java
lidavidm, line 1: nit: AvroProducer since everything else is named like that?

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/producers/Producer.java
```diff
@@ -0,0 +1,62 @@
...
+   * @param encoder avro encoder to write data
+   * @throws IOException on error
+   */
+  void produce(Encoder encoder) throws IOException;
+
+  /** Skip null value in the vector by setting reader position + 1. */
+  void skipNull();
+
+  /** Set the position to read value from vector. */
+  void setPosition(int index);
```
lidavidm, line 43: I think we've been trying to be future-proof and use `long` for indices now (since eventually once we support the new memory APIs we'll be able to address more than 2GiB of memory at once)

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/producers/BaseAvroProducer.java
```diff
@@ -0,0 +1,66 @@
...
+  public void close() throws Exception {
+    vector.close();
+  }
+
+  @Override
+  public boolean resetValueVector(T vector) {
+    this.vector = vector;
+    this.currentIndex = 0;
+    return true;
+  }
```
lidavidm, line 60: Hmm. Should resetting close the prior vector? If not, then maybe we also shouldn't be closing the vector here (i.e. it's the upper level code's responsibility to manage the vector lifetime and the producer is just temporarily holding the vector, not owning it)

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/producers/CompositeAvroProducer.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+  }
+
+  @Override
+  public void close() {
+    // clean up
+    try {
+      AutoCloseables.close(producers);
+    } catch (Exception e) {
+      throw new RuntimeException("Error occurs in close.", e);
```
lidavidm, line 66: Normally `close` is declared to throw `Exception` anyways - do we need to rewrap this in RuntimeException?

Otherwise use `AutoCloseables.closeNoChecked`

lidavidm (commented):
This looks reasonable to me overall, thanks.

- We should document exactly how we handle type conversions, especially around nulls
- I'm OK punting on more complicated types for now
- I'm wondering if we shouldn't just make the producer not own the vector at all (i.e. it's still the caller's responsibility to close the vector). I think this makes the most sense, e.g. if you have a file reader, the reader already owns the vector, so using it with the avro producer would mean a double-close (or you would have to explicitly unload the data and load it somewhere else)
- I'm OK punting on dictionaries for the time being too if you want

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/producers/BaseAvroProducer.java
martin-traverse: I think I agree with you that producers should not own / close the vectors. In which case, this becomes a non-issue! Are you happy for me to update on those lines? The producers will still own a reference, but we could get rid of all the close() methods and remove AutoClosable from the base interface. Would you like me to do that?

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/producers/CompositeAvroProducer.java
martin-traverse: Per above, if producers are not owning / closing the vectors this issue goes away. Otherwise yes this would make sense. Let me know, which you'd like me to do.

martin-traverse:
Thanks so much for all the comments! I've put questions in line, will wait for your answers on how to address those points.

In the meantime I will add the schema generator and producers for the narrow primitive types, as well as some top level doc comments describing the translation. I'll make a start on the test coverage as well....

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/producers/Producer.java
lidavidm: That sounds good to me.

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/producers/BaseAvroProducer.java
lidavidm: Yeah, let's drop `close()` and `AutoClosable`!

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/producers/BaseAvroProducer.java
```diff
@@ -0,0 +1,66 @@
...
+
+  @Override
+  public void skipNull() {
+    currentIndex++;
+  }
+
+  @Override
+  public void setPosition(int index) {
+    currentIndex = index;
```
wgtmac, line 47: Is there any restriction to the new `index`? Do we expect `index` < `currentIndex`?

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/producers/logical/AvroDecimalProducer.java
```diff
@@ -0,0 +1,78 @@
...
+  public static class BytesDecimalProducer extends AvroDecimalProducer {
+
+    private final byte[] reuseBytes;
+
+    /** Instantiate a BytesDecimalConsumer. */
+    public BytesDecimalProducer(DecimalVector vector) {
+      super(vector);
+      Preconditions.checkArgument(
+          vector.getTypeWidth() <= 16, "Decimal bytes length should <= 16.");
+      reuseBytes = new byte[vector.getTypeWidth()];
```
wgtmac, line 46: Is this correct? I don't think the type width is constant for each value. It might be similar to what you did in `AvroBytesProducer`.

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/producers/AvroNullableProducer.java
```diff
@@ -0,0 +1,77 @@
...
+  /** Instantiate a AvroNullableProducer. */
+  public AvroNullableProducer(Producer<T> delegate) {
+    super(delegate.getVector());
+    this.delegate = delegate;
+  }
+
+  @Override
+  public void produce(Encoder encoder) throws IOException {
+    if (vector.isNull(currentIndex)) {
+      encoder.writeInt(1);
```
wgtmac, line 40: Does it mean that you placed `[type, null]` as the union type? Is it a good convention to put `null` before `type` so the default value is null when reader type does not match the writer type?

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/ArrowToAvroUtils.java
```diff
@@ -0,0 +1,194 @@
...
+      case LIST:
+        ListVector listVector = (ListVector) vector;
+        FieldVector itemVector = listVector.getDataVector();
+        Producer<?> itemProducer = createProducer(itemVector, itemVector.getField().isNullable());
+        return new AvroArraysProducer(listVector, itemProducer);
+
+      case MAP:
+        MapVector mapVector = (MapVector) vector;
+        StructVector entryVector = (StructVector) mapVector.getDataVector();
+        VarCharVector keyVector = (VarCharVector) entryVector.getChildrenFromFields().get(0);
```
wgtmac, line 165: What if the key is not a string type?

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/producers/logical/AvroDecimalProducer.java
lidavidm: Decimal is fixed width byte (either 16 or 32)

wgtmac: Well, I think I have mixed up the requirement from Iceberg which says `Stored as fixed using the minimum number of bytes for the given precision.`

martin-traverse:
Hello. I have pushed a second iteration for review. Major updates are:

* Added schema generation, to convert Arrow List<Field> to Avro Schema (record schema). I added doc comments with the type mapping. Schema is built of ArrowType while physical producers use MinorType, that bugged me at first but on reflection probably seems correct?
* Updated the producers with a more complete set of types, including all widths of float / int types and logical types for all the date / time types (but not duration as of yet). Would appreciate a set of eyes on the timezone mapping for TZ aware timestamps! Per my reading of the Avro spec this is correct but worth double checking.

I have tried to make it so most Arrow types will map to an Avro type if there is a suitable target. The basic approach is:

* If a direct equivalent exists, use that
* Next, choose an equivalent with no loss of precision (e.g. small int -> int)
* Next, choose an equivalent with loss of precision but no loss of range (e.g. time nano -> time micros)
* Finally, choose an equivalent with the best match and error on overflow (e.g. unsigned long -> long)

I'm not sure about the last one, would it be better just to say we can't convert those types and throw an error up front?

Large types (for text, binary, list) are not supported yet, there will be some extra thought needed on getting very large buffers between the two APIs. Interval / duration also not done yet, Arrow Interval resembles Avro Duration but is not an exact match and Arrow duration doesn't really have an equivalent that I could see. View types also not included as of now.

I tried to move over to using long as the current row index, however a lot of the high-level APIs on the vector types use ints. For working directly with primitive types I was able to go to the underlying buffer (there were reasons to do that anyway in a few cases), but for the complex types that gets trickier because the type of the child vectors is not known. Is it ok to keep the interface using an int for current index for now? Most of the primitive producers are updated anyway, so the work to move over later would not be too much.

I've run out of time for today. Next I plan to do the tests - like the consumers, will be exhaustive across types, combinations of nullability etc with some edge cases. Then the last thing I'd like to come back to on this PR is enums / dict encoding, just in case that has any impact on the structure of things would be good to know while it's still easy to change things.

I have addressed some of the comments but not all, I will work through the rest but wanted to share this to check I'm still going in the right direction!

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/producers/BaseAvroProducer.java
martin-traverse: Done, close() and AutoClosable are gone  :-)

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/producers/CompositeAvroProducer.java
martin-traverse: Per above, close and AutoCloable are gone

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/producers/Producer.java
martin-traverse: Per my [comment below](https://github.com/apache/arrow-java/pull/638#issuecomment-2692815888), this proved more tricky than I first thought. In light of that, are you happy to stick with int for now?

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/producers/logical/AvroDecimalProducer.java
martin-traverse: Apologies, have removed this, I was mirroring consumers which need to handle both bytes and fixed inputs, but for producers we can always used fixed and the type width is always known.

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/producers/AvroNullableProducer.java
martin-traverse: I tried doing that, it broke the consumers and the data would not do round trip. In Avro this is considered as the difference between "nullable" and "optional", with "optional" being the semantics you are suggesting. I have kept it as nullable for now. I do think we will need to follow up with a PR on the consumers, particularly to address null handling. Nullable types get mangled in the schema round trip at the moment!

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/ArrowToAvroUtils.java
martin-traverse: Good point! I put a check in the schema generation but not here. In Avro keys have to be string so if that is not the case I have thrown an error. Please leave this open and I will fix - out of time for today!

lidavidm:
Sorry, I've gotten caught up in my main job - I'll try to circle back here soon

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/producers/logical/BaseTimestampTzProducer.java
```diff
@@ -0,0 +1,58 @@
...
+import java.time.ZoneId;
+import java.time.ZoneOffset;
+import org.apache.arrow.adapter.avro.producers.BaseAvroProducer;
+import org.apache.arrow.vector.TimeStampVector;
+import org.apache.avro.io.Encoder;
+
+abstract class BaseTimestampTzProducer<T extends TimeStampVector>
+    extends BaseAvroProducer<TimeStampVector> {
+
+  // Convert TZ values to UTC to encode Avro timestamp types
```
lidavidm, line 29: Arrow stores zoned timestamps in UTC already. The time zone associated with the type is purely used for display. So there should be no need for conversion.

https://github.com/apache/arrow/blob/601fde9033b5def7e667ffec70c0acacd1b9ad50/format/Schema.fbs#L284-L302

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/producers/AvroUint4Producer.java
```diff
@@ -0,0 +1,40 @@
...
+
+  /** Instantiate an AvroUint4Producer. */
+  public AvroUint4Producer(UInt4Vector vector) {
+    super(vector);
+  }
+
+  @Override
+  public void produce(Encoder encoder) throws IOException {
+    int unsigned = vector.getDataBuffer().getInt(currentIndex * (long) UInt4Vector.TYPE_WIDTH);
+    long signed = unsigned & 0xffffffffL;
```
lidavidm, line 36: Is this https://docs.oracle.com/javase/8/docs/api/java/lang/Integer.html#toUnsignedLong-int- ?

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/producers/AvroUint2Producer.java
```diff
@@ -0,0 +1,38 @@
...
+
+  /** Instantiate an AvroUint2Producer. */
+  public AvroUint2Producer(UInt2Vector vector) {
+    super(vector);
+  }
+
+  @Override
+  public void produce(Encoder encoder) throws IOException {
+    short unsigned = vector.getDataBuffer().getShort(currentIndex * (long) UInt2Vector.TYPE_WIDTH);
+    int signed = unsigned & 0xffff;
```
lidavidm, line 34: Same here, is Short#toUnsignedInt applicable?

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/ArrowToAvroUtils.java
```diff
@@ -0,0 +1,729 @@
...
+  public static final String GENERIC_RECORD_TYPE_NAME = "GenericRecord";
+
+  /**
+   * Create an Avro record schema for a given list of Arrow fields.
+   *
+   * <p>This method currently performs following type mapping for Avro data types to corresponding
+   * Arrow data types.
+   *
+   * <ul>
+   *   <li>ArrowType.Null --> NULL
```
lidavidm, line 122: not to nitpick the formatting but a table might be better? (also does Javadoc allow unclosed tags like this?)

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/ArrowToAvroUtils.java
```diff
@@ -0,0 +1,729 @@
...
+   *   <li>ArrowType.Struct --> record
+   *   <li>ArrowType.List --> array
+   *   <li>ArrowType.LargeList --> array
+   *   <li>ArrowType.FixedSizeList --> array
+   *   <li>ArrowType.Map --> map
+   *   <li>ArrowType.Union --> union
+   * </ul>
+   *
+   * <p>Nullable fields are represented as a union of [null | base-type]. Special treatment is given
+   * to nullability of unions - a union is considered nullable if the union field is nullable or any
```
lidavidm, line 154: FWIW, I think nullable unions aren't semantically valid - unions have no validity bitmap in Arrow. The only way to represent this is with a nullable child field

martin-traverse:
Thanks @lidavidm for all the comments - really helpful :-) I'll push an update in a few days to address these and also add the tests, will probably be after the weekend.

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/ArrowToAvroUtils.java
martin-traverse: I added a check for this, now it will throw illegal argument if the key type is not string (which is required by Avro)

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/producers/Producer.java
```diff
@@ -0,0 +1,53 @@
...
+
+  /**
+   * Reset the vector within producer.
+   *
+   * @return true if reset is successful, false if reset is not needed.
+   */
+  boolean resetValueVector(T vector);
```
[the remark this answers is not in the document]
martin-traverse: Nothing ever returns false and the value is never used. I have changed this to return void.

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/producers/AvroStructProducer.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+    super.setPosition(index);
+  }
+
+  @Override
+  @SuppressWarnings("unchecked")
+  public boolean resetValueVector(StructVector vector) {
+    for (int i = 0; i < delegates.length; i++) {
+      Producer<FieldVector> delegate = (Producer<FieldVector>) delegates[i];
+      delegate.resetValueVector(vector.getChildrenFromFields().get(i));
```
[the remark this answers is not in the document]
martin-traverse: Per above, nothing ever returns false and the value is never used, so I removed the return value.

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/producers/logical/AvroTimeNanoProducer.java
```diff
@@ -0,0 +1,46 @@
...
+  private static final long NANOS_PER_MICRO = 1000;
+
+  public AvroTimeNanoProducer(TimeNanoVector vector) {
+    super(vector);
+  }
+
+  @Override
+  public void produce(Encoder encoder) throws IOException {
+    long nanos = vector.getDataBuffer().getLong(currentIndex * (long) TimeNanoVector.TYPE_WIDTH);
+    long micros = nanos / NANOS_PER_MICRO;
```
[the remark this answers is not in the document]
martin-traverse: I don't think this will truncate since we are going nanos -> micros so there is more range available. There is loss of precision - I think that is acceptable, since Avro does not have a time-nanos type (at least not yet)!

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/producers/logical/AvroTimeMilliProducer.java
```diff
@@ -0,0 +1,34 @@
...
+import org.apache.arrow.adapter.avro.producers.AvroIntProducer;
+import org.apache.arrow.vector.TimeMilliVector;
+
+/**
+ * Producer that produces time (milliseconds) values from a {@link TimeMilliVector}, writes data to
+ * an Avro encoder.
+ */
+public class AvroTimeMilliProducer extends AvroIntProducer {
+
+  // Time in milliseconds stored as integer, matches Avro time-millis type
```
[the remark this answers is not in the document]
martin-traverse: I think time for seconds / millis are four byte width? [TimeMilliVector](https://github.com/apache/arrow-java/blob/48a20ba91605ebb44df845671782acd8f869bd4b/vector/src/main/java/org/apache/arrow/vector/TimeMilliVector.java#L39). So shouldn't be a range issue for this one.

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/producers/logical/BaseTimestampTzProducer.java
martin-traverse: This is great! I took out all the timezone conversion logic, the TZ aware producers are much simpler now :-)

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/producers/AvroUint4Producer.java
martin-traverse: Thanks, updated the code to use that method.

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/producers/AvroUint2Producer.java
martin-traverse: Thanks, updated the code to use that method.

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/ArrowToAvroUtils.java
martin-traverse: I've converted the type mapping info into and HTML table - shows up nicely in the IDE auto complete :-)

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/ArrowToAvroUtils.java
martin-traverse: I checked the spec and yes you are correct!  I have updated the logic accordingly

lidavidm:
I'm fine splitting things into other PRs if it makes your life easier.

It is known that the Java Union implementation is wrong, unfortunately. See https://github.com/apache/arrow-java/issues/108

If you want to punt on Union for now that's fine by me.

I'm a bit backlogged and will try to re-review when I get a chance but that may not be for a week or two, sorry.

martin-traverse:
Thanks @lidavidm  - good to know it is not just me going crazy! I have removed support of data production for unions for now. Also rebased on master.

Totally understand the time pressure. I might work on some of the other pieces in the meantime as separate PRs if that's OK, to have a couple of bits ready for you to look at :-)

Also read the thread on Unions. My company uses Arrow Java heavily in our product - time is a big issue for us as well but if I can find some I'll try to raise a few fixes that we're interested in.

martin-traverse:
Hi @lidavidm  - There were two test failures in CI relating to nano precision time values that didn't show up in my local machine. I have pushed a fix.

martin-traverse:
CI looks good now - just the link missing on the PR to issue #615, sorry I don't have permission to add that!

martin-traverse:
I think the PR check hit [this issue](https://github.com/apache/arrow-java/pull/673) - could you add the milestone to the issue manually for now?

lidavidm:
Please ignore the check :grimacing:

martin-traverse:
> Please ignore the check 😬

Ok great - in that case I think the CI is done :)

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/producers/AvroListProducer.java
```diff
@@ -0,0 +1,73 @@
...
+
+    encoder.writeArrayEnd();
+    currentIndex++;
+  }
+
+  // Do not override skipNull(), delegate will not have an entry if the list is null
+
+  @Override
+  public void setPosition(int index) {
+    if (index < 0 || index > vector.getValueCount()) {
```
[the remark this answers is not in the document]
lidavidm: Ah, I see your comment below now. It might be worth noting this somewhere (perhaps the docstring of the base class?)

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/ArrowToAvroUtils.java
lidavidm: Can we update the docstring too?

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/ArrowToAvroUtils.java
```diff
@@ -0,0 +1,718 @@
...
+
+      case Date:
+        return builder.intBuilder().prop("logicalType", "date").endInt().noDefault();
+
+      case Time:
+        ArrowType.Time timeType = (ArrowType.Time) field.getType();
+        if ((timeType.getUnit() == TimeUnit.SECOND || timeType.getUnit() == TimeUnit.MILLISECOND)) {
+          return builder.intBuilder().prop("logicalType", "time-millis").endInt().noDefault();
+        } else {
+          // All other time types (sec, micro, nano) are encoded as time-micros (LONG)
```
lidavidm, line 415: nit: comment is out of sync with code

lidavidm (approved):

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/producers/AvroListProducer.java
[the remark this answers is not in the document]
martin-traverse: I have added a doc comment on the base class as you suggested

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/ArrowToAvroUtils.java
martin-traverse: I have updated the comment on unions. Also updated the comment for nullable types, which are encoded as [ base-type | null ], per the discussion above.

martin-traverse:
Hi - I think I have addressed all the new comments. Also found an issue in the fixed list producer - I updated the test cases to show this up and included a fix. Fairly sure all the complex type producers are correct now but would be worth a set of eyes!! The four I have are list, fixed size list, map and struct.

Please let me know if there is anything else needs addressing.

martin-traverse:
This last update uses Avro's logical type classes instead of setting the raw schema props. The output to disk is the same, but if you want to use the generated Avro Schema directly these structures should be available. Needed for round-trip testing.

lidavidm:
The proposal looks good to me. Sorry, I will try to give a hopefully final review here this week - as you have seen I tend to be scattered about so I appreciate your patience

lidavidm (approved):

lidavidm:
Ok well these CI failures aren't related to the PR, merging...

martin-traverse:
Hi @lidavidm  - Thanks so much for merging, I'm really happy to get this first piece in :-) Have started work on the second bit, once the main pieces are ready I will raise a new PR.

I'm  snowed under too so may have to put it down and come back a week later etc., very happy to work in async mode!

PR #650, opened 2025-03-04
MINOR: Upgrade to GitHub actions/cache 4.2.2

jbonofre:
Upgrade to GitHub actions/cache 4.2.2 as previous versions are deprecated.

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -42,7 +42,7 @@ jobs:
         with:
           python-version: '3.x'
       - name: pre-commit (cache)
-        uses: actions/cache@v4
+        uses: actions/cache@d4323d4df104b026a6aa633fdb11d772146be0bf # v4.2.2
         with:
           path: ~/.cache/pre-commit
           key: pre-commit-${{ hashFiles('.pre-commit-config.yaml') }}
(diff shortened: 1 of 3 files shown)


assignUser (commented):
-0.5
It is best practice to pin actions to SHA in workflows that run in an elevated context (i.e. not pull_request) and use secrets or have elevated permissions (anything not `contents:read`). So at least for those I would stick with pinned actions.

I also don't see the downside of pinning things, as I have seen actions/* break on minor updates in the past. Except maybe the work to merge dependabot PRs.

Also to be pedantic this touches more than 2 files which means it's not a minor PR ;

jbonofre:
@assignUser ok. Fair enough. Let me update this PR to fix the cache version which is deprecated (and blocking all PRs right now).

assignUser:
Ah I see the deprecation for anything <4.2. I forgot about that, IIRC they wanted to do that last year but pushed it out after some backlash. Somehow the dependabot PR must have been closed which makes it ignore that version. Though I can't find one in the PRs... hm, maybe broken through the transition or something.

lidavidm:
FWIW, Dependabot appears to understand the pinning convention, so at least it shouldn't be extra work to maintain it (e.g. https://github.com/apache/arrow-java/pull/635)

assignUser:
> FWIW, Dependabot appears to understand the pinning convention, so at least it shouldn't be extra work to maintain it (e.g. #635)

Yeah, otherwise it would be a real drain...

jbonofre:
@assignUser @lidavidm I updated this PR to upgrade to GitHub actions/cache 4.2.2 (using the SHA1).

PR #652, opened 2025-03-05
MINOR: Update versions in LICENSE/NOTICE

jbonofre:
Align versions in `LICENSE`/`NOTICE` in regards of the recent dependency updates.

### DIFF
--- a/flight/flight-sql-jdbc-driver/src/shade/NOTICE.txt
+++ b/flight/flight-sql-jdbc-driver/src/shade/NOTICE.txt
@@ -73,7 +73,7 @@ This product includes gRPC 1.70.0, with the following in its NOTICE:
 
 ---------------------------------------------------
 
-This product includes Netty 4.1.117.Final, with the following in its NOTICE:
+This product includes Netty 4.1.119.Final, with the following in its NOTICE:
 
 |                             The Netty Project
 |                             =================
--- a/flight/flight-integration-tests/src/shade/LICENSE.txt
+++ b/flight/flight-integration-tests/src/shade/LICENSE.txt
@@ -203,7 +203,7 @@
 
 --------------------------------------------------------------------------------
 
-This binary artifact contains Jackson 2.18.2.
+This binary artifact contains Jackson 2.18.3.
 
 Home page: https://github.com/FasterXML/jackson
 License: https://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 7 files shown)


lidavidm (approved):

flight/flight-sql-jdbc-driver/src/shade/NOTICE.txt
```diff
@@ -73,7 +73,7 @@ This product includes gRPC 1.70.0, with the following in its NOTICE:
 
 ---------------------------------------------------
 
-This product includes Netty 4.1.117.Final, with the following in its NOTICE:
+This product includes Netty 4.1.119.Final, with the following in its NOTICE:
```
kou, line 76: It seems that https://github.com/apache/arrow-java/tree/main/flight/flight-sql-jdbc-driver doesn't include copy of part of Netty. Do we need this? (Or should we change this to "This binary artifact contains ..."?)

LICENSE.txt
```diff
@@ -202,7 +202,7 @@
    limitations under the License.
 
 --------------------------------------------------------------------------------
-This product includes code from Netty 4.1.117.Final:
+This product includes code from Netty 4.1.119.Final:
```
kou, line 205: Do we need to update this?

I think that we should use the version when we copied.

jbonofre: Yes that's correct. My bad: my script was too aggressive 😀
Let me fix that.

flight/flight-sql-jdbc-driver/src/shade/NOTICE.txt
jbonofre: https://github.com/apache/arrow-java/blob/main/flight/flight-sql-jdbc-driver/pom.xml#L120

kou: I'm not familiar with Java but I thought `flight-sql-jdbc-driver-19.0.0-SNAPSHOT.jar` is a binary artifact not a source archive. (I thought that https://repository.apache.org/content/repositories/staging/org/apache/arrow/flight-sql-jdbc-driver/18.2.0/flight-sql-jdbc-driver-18.2.0-sources.jar is a source archive for flight-sql-jdbc-driver.)

We use "This binary artifact contains ..." in other files for binary artifact and we use "This product includes ..." for copied sources. e.g.: https://github.com/apache/arrow-java/blob/cf8b2521ca43c862784b67d32af94861dea1d388/LICENSE.txt#L205

So I thought:
* We don't need this for source archive?
* We should use "This binary artifact ..."?

BTW, I noticed that this isn't used for `-sources.jar`! This is used only for `.jar`, right?

LICENSE.txt
kou: Wow! You have a script for this!? Could you share it?

flight/flight-sql-jdbc-driver/src/shade/NOTICE.txt
jbonofre: We have to ship LICENSE and NOTICE in the jar. That's why these files are in the shaded folder. And LICENSE has to list of dependencies shaded/shipped in the distributed artifact.

LICENSE.txt
jbonofre: Oh yes. I will create a PR to add the script and maybe include in CI.

kou (approved):
+1

Issue #655, opened 2025-03-07
Failure in UnionReader.read after DecimalVector promotion to UnionVector


lriggs:
### Describe the bug, including details regarding any error messages, version, and platform.

When a DecimalVector is promoted to a UnionVector via a PromotableWriter, the UnionVector will have the decimal vector in it's internal struct vector, but the decimalVector field will not be set. If UnionReader.read is then used to read from the UnionVector, it will fail when it tries to read one of the promoted decimal values, due to decimalVector being null, and the exact decimal type not being provided.  This failure is unnecessary though as we have a pre-existing decimal vector, the caller just does not know the exact type - and it shouldn't be required to.

PR #656, opened 2025-03-07
GH-655: Failure in UnionReader.read after DecimalVector promotion to UnionVector

lriggs:
When a DecimalVector is promoted to a UnionVector via a PromotableWriter, the UnionVector will have the decimal vector in it's internal struct vector, but the decimalVector field will not be set. If UnionReader.read is then used to read from the UnionVector, it will fail when it tries to read one of the promoted decimal values, due to decimalVector being null, and the exact decimal type not being provided.  This failure is unnecessary though as we have a pre-existing decimal vector, the caller just does not know the exact type - and it shouldn't be required to.

The change here is to check for a pre-existing decimal vector in the internal struct when getDecimalVector() is called.  If one exists, set the decimalVector field and return.  Otherwise, if none exists, throw the exception.

## What's Changed

Updated UnionVector to handle this case.

Closes GH-655

### DIFF
--- a/vector/src/main/codegen/templates/UnionVector.java
+++ b/vector/src/main/codegen/templates/UnionVector.java
@@ -280,7 +280,10 @@ public class UnionVector extends AbstractContainerVector implements FieldVector 
   <#if minor.class?starts_with("Decimal") || is_timestamp_tz(minor.class) || minor.class == "Duration" || minor.class == "FixedSizeBinary">
   public ${name}Vector get${name}Vector() {
     if (${uncappedName}Vector == null) {
-      throw new IllegalArgumentException("No ${name} present. Provide ArrowType argument to create a new vector");
+      ${uncappedName}Vector = internalStruct.getChild(fieldName(MinorType.${name?upper_case}), ${name}Vector.class);
+      if (${uncappedName}Vector == null) {
+        throw new IllegalArgumentException("No ${name} present. Provide ArrowType argument to create a new vector");
+      }
     }
     return ${uncappedName}Vector;
   }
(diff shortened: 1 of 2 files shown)


vector/src/main/codegen/templates/UnionVector.java
```diff
@@ -306,6 +307,9 @@ public StructVector getStruct() {
 
   public ${name}Vector get${name}Vector(String name) {
     if (${uncappedName}Vector == null) {
+      ${uncappedName}Vector = internalStruct.getChild(fieldName(MinorType.${name?upper_case}), ${name}Vector.class);
+      if (${uncappedName}Vector == null) {
+        throw new IllegalArgumentException("No ${uncappedName} present. Provide ArrowType argument to create a new vector");
```
lidavidm, line 312: Is it just me or is there no closing brace here?

Shouldn't we only add this new check for parameterized/nested types?

jbonofre: From a syntax standpoint (brace), it looks good to me.

vector/src/main/codegen/templates/UnionVector.java
```diff
@@ -306,6 +307,9 @@ public StructVector getStruct() {
 
   public ${name}Vector get${name}Vector(String name) {
     if (${uncappedName}Vector == null) {
+      ${uncappedName}Vector = internalStruct.getChild(fieldName(MinorType.${name?upper_case}), ${name}Vector.class);
```
jbonofre, line 310: Should we apply to only child, or also on parameterized types ?

vector/src/main/codegen/templates/UnionVector.java
lriggs: I pushed a new commit where this code is moved slightly, had forgotten that this one was updated in our repo.

vector/src/main/codegen/templates/UnionVector.java
lriggs: I'm not sure, I'm submitting a change someone else made in our repo. Would it be any different if used the parameterized type instead?

lidavidm (approved):

lidavidm:
It seems there's a syntax error still

lriggs:
I fixed the syntax error but not sure whats causing the latest Dev PR task error:
Linked issue: 655
failed to update https://github.com/apache/arrow-java/issues/655: GraphQL: Resource not accessible by integration (updateIssue)

lidavidm:
That's separate (I think I need to grant that step permissions to update issues, too)

lidavidm:
Are you able to `mvn spotless:apply` and commit the result? (Or would it be ok for me to push a commit here to fix up things?)

lidavidm:
Thanks!

PR #658, opened 2025-03-07
GH-657: Set Jna library loading symbol visibility

### LINKED ISSUE #657: [JNI][Gandiva] LLVM generated code can't access some system functions on certain platforms

When linux loads libraries there are different options for symbols visibility like: RTLD_LOCAL, RTLD_GLOBAL. Default in Java is RTLD_LOCAL which makes all symbols inside gandiva library only visible to the gandiva library. 

The problem is when llvm compiles code and loads it into Java process that code won't be able to find any symbols imported by gandiva. One example is 'sinhl' from libgcc which causes sinh function to fail when it is present in SQL query executed by Dremio using Gandiva engine. By importing jna package we can control how to load gandiva library and set Global symbol visibility.

lriggs:
## What's Changed

Set the Lazy and Global flags when loading the Gandiva library for jni so that system functions are visible to the compiled llvm code.

Closes #657 and https://github.com/apache/arrow/issues/40839.

### DIFF
--- a/gandiva/src/main/java/org/apache/arrow/gandiva/evaluator/JniLoader.java
+++ b/gandiva/src/main/java/org/apache/arrow/gandiva/evaluator/JniLoader.java
@@ -23,6 +23,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.nio.file.Files;
 import java.nio.file.StandardCopyOption;
+import java.util.Collections;
 import java.util.Locale;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
@@ -31,6 +32,9 @@ import org.apache.arrow.gandiva.exceptions.GandivaException;
 /** This class handles loading of the jni library, and acts as a bridge for the native functions. */
 class JniLoader {
   private static final String LIBRARY_NAME = "gandiva_jni";
+  
+  private static final int RTLD_GLOBAL = 0x00100;
+  private static final int RTLD_LAZY = 0x00001; 
 
   private static volatile JniLoader INSTANCE;
   private static volatile long defaultConfiguration = 0L;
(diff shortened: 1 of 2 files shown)


gandiva/src/main/java/org/apache/arrow/gandiva/evaluator/JniLoader.java
```diff
@@ -69,6 +75,9 @@ private static void loadGandivaLibraryFromJar(final String tmpDir)
     final String libraryToLoad =
         LIBRARY_NAME + "/" + getNormalizedArch() + "/" + System.mapLibraryName(LIBRARY_NAME);
     final File libraryFile = moveFileFromJarToTemp(tmpDir, libraryToLoad, LIBRARY_NAME);
+    NativeLibrary.getInstance(
+        libraryFile.getAbsolutePath(),
+        Collections.singletonMap(Library.OPTION_OPEN_FLAGS, new Integer(RTLD_LAZY | RTLD_GLOBAL)));
     System.load(libraryFile.getAbsolutePath());
```
lidavidm, line 81: Is this still necessary if we're using the JNA version?

gandiva/src/main/java/org/apache/arrow/gandiva/evaluator/JniLoader.java
```diff
@@ -69,6 +75,9 @@ private static void loadGandivaLibraryFromJar(final String tmpDir)
     final String libraryToLoad =
         LIBRARY_NAME + "/" + getNormalizedArch() + "/" + System.mapLibraryName(LIBRARY_NAME);
     final File libraryFile = moveFileFromJarToTemp(tmpDir, libraryToLoad, LIBRARY_NAME);
+    NativeLibrary.getInstance(
```
lidavidm, line 78: Doesn't the NativeLibrary instance need to be disposed somewhere?

gandiva/src/main/java/org/apache/arrow/gandiva/evaluator/JniLoader.java
jbonofre: I don't think this is strictly required with JNA imho.

gandiva/src/main/java/org/apache/arrow/gandiva/evaluator/JniLoader.java
jbonofre: Why not also leveraging `java.library.path` system prop ?

PR #660, opened 2025-03-07
GH-81: [Flight] Expose gRPC in Flight client builder

### LINKED ISSUE #81: Connect timeout for Flight RPC client in Java implementation
### Describe the usage question you have. Please include as many useful details as  possible.

I didn't manage to find a way to configure connect timeout when setting up Flight client. However this is possible for example in C# implementation by passing respective option (`GrpcChannelOptions`) when configuring `GrpcChannel` (not accessible in Java code). 
Am I correct and there is no way to control connect timeout or do I miss something? 

### Component(s)

Java

lidavidm:
## What's Changed

Expose the internal gRPC channel builder in the FlightClient so that applications can build on top of it without having to replicate all the special Flight logic. This makes it much easier to apply some gRPC tweaks without duplicating Arrow code.

Closes #81.


jbonofre (approved):
I see this is extracted from the other PR. It looks good to me. Thanks @lidavidm !

PR #662, opened 2025-03-10
GH-661: [Flight] JDBC: Cache failed locations

### LINKED ISSUE #661: [Flight] JDBC driver: cache failed connection attempts

Splitting up https://github.com/apache/arrow-java/issues/494 to make review more manageable. 

If a Flight SQL server returns a set of locations to try, and one of the locations is never reachable, we currently wait the full connect timeout on every attempt. This is rather painful. As a stepping stone to full caching of clients (https://github.com/apache/arrow-java/issues/157), we can cache failures and avoid trying those locations first on subsequent queries (we can still try them last in case other locations fail).

lidavidm:
## What's Changed

Instead of repeatedly trying to access locations which have failed before, cache them and move them to the end of the list of locations to attempt on subsequent accesses. This provides a better user experience for Flight SQL services that return multiple locations in the expectation that the user may not be able to access some of them (but would have a better experience if they are).

Requires #494.

Closes #661.

### DIFF
--- a/flight/flight-sql-jdbc-core/pom.xml
+++ b/flight/flight-sql-jdbc-core/pom.xml
@@ -147,6 +147,12 @@ under the License.
       <groupId>org.checkerframework</groupId>
       <artifactId>checker-qual</artifactId>
     </dependency>
+
+    <dependency>
+      <groupId>com.github.ben-manes.caffeine</groupId>
+      <artifactId>caffeine</artifactId>
+      <version>3.1.8</version>
+    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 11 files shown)


flight/flight-sql-jdbc-core/pom.xml
```diff
@@ -147,6 +147,12 @@ under the License.
       <groupId>org.checkerframework</groupId>
       <artifactId>checker-qual</artifactId>
     </dependency>
+
+    <dependency>
+      <groupId>com.github.ben-manes.caffeine</groupId>
+      <artifactId>caffeine</artifactId>
```
jbonofre, line 153: Do we really need caffeine here ? I don't see obvious usage in other part of this PR.

lidavidm: It is used in FlightClientCache.java.

jbonofre: Oh yeah, I see it now, sorry for the noise 😄

PR #670, opened 2025-03-10
MINOR: Add check-shebang-scripts-are-executable

lidavidm:
## What's Changed

Make sure that new scripts committed are actually executable, because I always forget.


jbonofre (approved):

kou:
BTW, this may help you:

```elisp
(add-hook 'after-save-hook 'executable-make-buffer-file-executable-if-script-p)
```

PR #671, opened 2025-03-10
Add jbonofre in CODEOWNERS and remove from collaborators list

jbonofre:
As jbonofre has been elected as committers, collaborators entry is not required anymore. And happy to be in `CODEOWNERS` to help on review.


jbonofre:
@kou thanks again for everything !

PR #689, opened 2025-03-24
MINOR: Fix macOS JNI build

lidavidm:
## What's Changed

Update a sanity check in CI that's incorrectly failing the macOS JNI builds.


jbonofre (approved):

lidavidm:
Hmm...why are we depending on libz3, actually...

lidavidm:
Ah, it's because Gandiva (so then probably LLVM) now uses Z3.

jbonofre:
> Hmm...why are we depending on libz3, actually...

Afair, gandiva upgraded to Z3, that's probably the cause 😄

PR #693, opened 2025-03-27
Preserve nullability information while transfering DecimalVector and Decimal256Vector

### LINKED ISSUE #692: [arrow-vector] Zero-copy transfer of DecimalVector and Decimal256Vector is losing nullability information

In the main branch, `DecimalVector` and `Decimal256Vector` are losing nullability information during a zero-copy transfer. It currently **forces** a nullable `FieldType` onto the "to" `ValueVector`.

The deficiency in the code is in `DecimalVector.TransferImpl` class when trying to get a transfer  `getTransferPair(String, BufferAllocator)` creating a `DecimalVector` with a nullable `FieldType` (although the "from" ValueVector's FieldType is non-nullable):
https://github.com/apache/arrow-java/blob/48a20ba91605ebb44df845671782acd8f869bd4b/vector/src/main/java/org/apache/arrow/vector/DecimalVector.java#L563-L569

Note that all the other (at least the primitive) `ValueVector`s are transferred without losing nullability information by using the "from" `ValueVector`'s field. E.g.:
https://github.com/apache/arrow-java/blob/48a20ba91605ebb44df845671782acd8f869bd4b/vector/src/main/java/org/apache/arrow/vector/BigIntVector.java#L319-L324

---

To **reproduce** the issue: add the following test in `src/test/java/org/apache/arrow/vector/TestDecimalVector.java` to see it fail.
```java
  @Test
  public void testGetTransferPairWithoutFieldNonNullable() {
    final FieldType decimalNonNullableType =
        new FieldType(false, new ArrowType.Decimal(10, scale), null);
    final DecimalVector fromVector =
        new DecimalVector("decimal", decimalNonNullableType, allocator);
    final TransferPair transferPair =
        fromVector.getTransferPair(fromVector.getField().getName(), allocator);
    final DecimalVector toVector = (DecimalVector) transferPair.getTo();

    // These asserts fail
    assertSame(fromVector.getField().isNullable(), toVector.getField().isNullable());
    assertSame(fromVector.getField().getFieldType(), toVector.getField().getFieldType());
  }
```

bodduv:
## What's Changed

This PR proposes to use the "from" `ValueVector`'s field while transferring `DecimalVector` and `Decimal256Vector` to preserve nullability information. This is to have similar behavior with all the other primitive value vectors. Note that the `FieldType` of the value vector has nullability information.

Closes #692 .

### DIFF
--- a/vector/src/main/java/org/apache/arrow/vector/Decimal256Vector.java
+++ b/vector/src/main/java/org/apache/arrow/vector/Decimal256Vector.java
@@ -566,9 +566,7 @@ public final class Decimal256Vector extends BaseFixedWidthVector
     Decimal256Vector to;
 
     public TransferImpl(String ref, BufferAllocator allocator) {
-      to =
-          new Decimal256Vector(
-              ref, allocator, Decimal256Vector.this.precision, Decimal256Vector.this.scale);
+      to = new Decimal256Vector(ref, Decimal256Vector.this.field.getFieldType(), allocator);
     }
 
     public TransferImpl(Field field, BufferAllocator allocator) {
--- a/vector/src/main/java/org/apache/arrow/vector/DecimalVector.java
+++ b/vector/src/main/java/org/apache/arrow/vector/DecimalVector.java
@@ -564,8 +564,7 @@ public final class DecimalVector extends BaseFixedWidthVector
     DecimalVector to;
 
     public TransferImpl(String ref, BufferAllocator allocator) {
-      to =
-          new DecimalVector(ref, allocator, DecimalVector.this.precision, DecimalVector.this.scale);
+      to = new DecimalVector(ref, DecimalVector.this.field.getFieldType(), allocator);
     }
 
     public TransferImpl(Field field, BufferAllocator allocator) {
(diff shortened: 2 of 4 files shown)


vector/src/main/java/org/apache/arrow/vector/Decimal256Vector.java
```diff
@@ -566,9 +566,7 @@ private class TransferImpl implements TransferPair {
     Decimal256Vector to;
 
     public TransferImpl(String ref, BufferAllocator allocator) {
-      to =
-          new Decimal256Vector(
-              ref, allocator, Decimal256Vector.this.precision, Decimal256Vector.this.scale);
+      to = new Decimal256Vector(ref, Decimal256Vector.this.field.getFieldType(), allocator);
```
jbonofre, line 569: Should we not test with `field`/`getFiedlType()` is not null and fallback to the vector precision if null ?

vector/src/main/java/org/apache/arrow/vector/DecimalVector.java
```diff
@@ -564,8 +564,7 @@ private class TransferImpl implements TransferPair {
     DecimalVector to;
 
     public TransferImpl(String ref, BufferAllocator allocator) {
-      to =
-          new DecimalVector(ref, allocator, DecimalVector.this.precision, DecimalVector.this.scale);
+      to = new DecimalVector(ref, DecimalVector.this.field.getFieldType(), allocator);
```
jbonofre, line 567: Same here. AFAIR, `field` and `getFieldType()` can be `null` (even if it should not happen, nothing prevent that in the code).

jbonofre (approved):
LGTM.

@lidavidm @kou do you mind to take a look ? If good, I propose to merge.

lidavidm (approved):

Issue #88, opened 2024-05-16
[Java][C] sliced RecordBatch offset info is lost when imported from c-data


hellishfire:
### Describe the bug, including details regarding any error messages, version, and platform.

Reproduced on latest arrow release (16.0)

When importing a sliced RecordBatch from c to java

On c side:
```
auto sliced_record_batch = original_record_batch->Slice(/*offset=*/8, /*length=*/2);
arrow::ExportRecordBatch(sliced_record_batch, arrow_array_ptr);
```

On java side:
```
ArrowArray arrowArray = ArrowArray.allocateNew(allocator);
Data.importIntoVectorSchemaRoot(allocator, arrowArray, vectorSchemaRoot, null);
```

The imported vectorSchemaRoot maintains the correct length(which is 2), but the offset info (which is 8) is not respected, hence the content of the imported vectorSchemaRoot points to the first 2 rows of the original_record_batch, while the desired content is sliced_record_batch.

I'm not familiar with arrow code, but it seems that the offset info is actually present in org.apache.arrow.c.ArrowArray.Snapshot, but org.apache.arrow.c.ArrayImporter ignores the offset in org.apache.arrow.c.ArrayImporter.doImport(ArrowArray.Snapshot)

### Component(s)

Java

PR #705, opened 2025-04-07
GH-704: Fix initialization of offset buffer when exporting VarChar vectors through C Data Interface

### LINKED ISSUE #704: Offset buffer contains garbled data when exporting empty string arrays from Java

https://github.com/apache/arrow/pull/40043 fixes exporting empty string arrays by exporting a non-empty offset buffer. However, the offset buffer is not properly initialized. The first byte of the offset buffer may not be 0 and causes weird errors on the importing side. See https://github.com/apache/datafusion-comet/issues/1615 for a issue of datafusion-comet caused by this.

This may also relates to https://github.com/apache/arrow-java/issues/88. https://github.com/apache/arrow/pull/40043 introduces a bug that corrupts the offset buffer before reading.

Kontinuation:
## What's Changed

This patch fixes the initialization of offset buffers when exporting variable width arrays through Arrow C Data Interface. The original code incorrectly mess up with the member `this.offsetBuffer` while we should actually initialize the newly allocated offsetBuffer. I think the diff itself will be quite self-explanatory.

Closes #704 and probably #88 .

### DIFF
--- a/c/src/test/java/org/apache/arrow/c/RoundtripTest.java
+++ b/c/src/test/java/org/apache/arrow/c/RoundtripTest.java
@@ -79,7 +79,6 @@ import org.apache.arrow.vector.UInt8Vector;
 import org.apache.arrow.vector.ValueVector;
 import org.apache.arrow.vector.VarBinaryVector;
 import org.apache.arrow.vector.VarCharVector;
-import org.apache.arrow.vector.VariableWidthVector;
 import org.apache.arrow.vector.VectorSchemaRoot;
 import org.apache.arrow.vector.ViewVarBinaryVector;
 import org.apache.arrow.vector.ViewVarCharVector;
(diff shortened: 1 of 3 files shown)


c/src/test/java/org/apache/arrow/c/RoundtripTest.java
```diff
@@ -77,6 +79,7 @@
 import org.apache.arrow.vector.ValueVector;
 import org.apache.arrow.vector.VarBinaryVector;
 import org.apache.arrow.vector.VarCharVector;
+import org.apache.arrow.vector.VariableWidthVector;
```
lidavidm, line 82: Unintentional import? This is failing the build

lidavidm (approved):

c/src/test/java/org/apache/arrow/c/RoundtripTest.java
Kontinuation: Fixed. Other test failures seems to be unrelated to this PR.

lidavidm (approved):

PR #707, opened 2025-04-10
GH-709: Correct length calculation of value buffers of variable-sized arrays

### LINKED ISSUE #74: [Java] Handle `offset` field from `ArrowArray` when `BufferImportTypeVisitor` imports offset buffer

This bug is found during debugging the issue https://github.com/apache/datafusion-comet/issues/540.

We found that some string arrays' offsets are out of the range of their value buffer. I.e., a string array's value buffer is only 147456 bytes, but the offsets of the last string is (294894, 294912).

The string is output from DataFusion aggregation operator `AggregateExec`. When producing the output batch, the operator will possibly slice the output batch if it is larger than a configured size. The slice of a string array in arrow-rs, keeps original value buffer and moves the pointer of offset buffer so it is a zero-copy slice.

During `importOffsets` call in `BufferImportTypeVisitor`, the slice of offset can be imported correctly as it uses the moved pointer and calculates the offset buffer correctly.

But when it goes to import value buffer, it calculates the capacity of it by using the difference between imported last offset and first offset. Because the imported offsets are from the slice, the calculated capacity is only for the certain slice of the value buffer.

For example, the original string array's value buffer is 346536 bytes, last offset is 346536. We take a slice of 8192 strings from it. The slice array's last offset is 294912 but the value buffer is the same (346536 bytes).

When `BufferImportTypeVisitor` imports the slice, the imported offsets are [147456, ..., 294912]. It calculates the length of value buffer is `294912 - 147456 = 147456`. But actually the length of value buffer is 346536.

Obviously the offsets are now out of range of the incorrect value buffer size 147456.

To be clear, the source of the issue comes from https://github.com/apache/arrow-rs/issues/5896 where it exports moved pointer of offer buffer of a slice of string array. Instead, we should use `offset` field in `ArrowArray` for it. We are going to fix it in arrow-rs for that.

But actually seems either ArrayImporter or `BufferImportTypeVisitor` also doesn't consider `offset` in `ArrowArray`.

### Component(s)

Java

### LINKED ISSUE #709: Accessing values from imported 'C data interface' array can result in out of bounds reads

Apache Datafusion creates variable-sized binary arrays whose first offset value into the value buffer is not `0`. When importing these into the Java implementation via the C data interface, `BufferImportTypeVisitor` calculates the size of the value buffer as `offsets[<last>] - offsets[0]`. In the case where `offsets[0] != 0` this results in a too small value for the capacity of the value buffer. When attempting to access the last value in the value buffer, this results in an out of bounds exception

pepijnve:
## What's Changed

For variable-size binary layout arrays, BufferImportTypeVisitor currently derives the length of the value buffer by calculating the difference between the last and first offset. When the first offset is not zero, this is actually incorrect and leads to out of bounds errors when attempting to read values from the imported array.

Instead, BufferImportTypeVisitor should simply use the last offset value as the length of the value buffer. This PR makes that change.

Just FYI, I bumped into this issue when attempting to import an array originating from DataFusion. A test query of the form `SELECT column1 FROM VALUES ('a'), ('b'), ('c'), ('d') LIMIT 2 OFFSET 1;` returns a slice of the full set of values. The values buffer contains all the original values, and the offsets buffer contains 1 and 2 as values to handle the offset from the query.

Closes #709 .

### DIFF
--- a/c/src/main/java/org/apache/arrow/c/BufferImportTypeVisitor.java
+++ b/c/src/main/java/org/apache/arrow/c/BufferImportTypeVisitor.java
@@ -228,9 +228,8 @@ class BufferImportTypeVisitor implements ArrowType.ArrowTypeVisitor<List<ArrowBu
           type,
           start,
           end);
-      final int len = end - start;
       offsets.getReferenceManager().retain();
-      return Arrays.asList(maybeImportBitmap(type), offsets, importData(type, len));
+      return Arrays.asList(maybeImportBitmap(type), offsets, importData(type, end));
     }
   }
 
@@ -279,9 +278,8 @@ class BufferImportTypeVisitor implements ArrowType.ArrowTypeVisitor<List<ArrowBu
           type,
           start,
           end);
-      final long len = end - start;
       offsets.getReferenceManager().retain();
-      return Arrays.asList(maybeImportBitmap(type), offsets, importData(type, len));
+      return Arrays.asList(maybeImportBitmap(type), offsets, importData(type, end));
     }
   }
 
@@ -296,9 +294,8 @@ class BufferImportTypeVisitor implements ArrowType.ArrowTypeVisitor<List<ArrowBu
           type,
           start,
           end);
-      final int len = end - start;
       offsets.getReferenceManager().retain();
-      return Arrays.asList(maybeImportBitmap(type), offsets, importData(type, len));
+      return Arrays.asList(maybeImportBitmap(type), offsets, importData(type, end));
     }
   }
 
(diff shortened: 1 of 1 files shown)


lidavidm:
arrow-java doesn't support slicing in general, so for offset-based arrays, other code downstream may not work properly if the first offset is nonzero. I think the longer term fix is to either detect this and copy or decide to properly implement slicing. That said fixing an out-of-bounds is always good.

lidavidm (approved):
LGTM.

Do you mind opening a new issue to link this PR to? The original issue has more discussion. In particular I think the cross-language integration tests need to be improved to cover this case

pepijnve:
I've created a new issue and update the summary and description of this PR.

PR #711, opened 2025-04-14
MINOR: Don't uninstall Flatbuffers in CI

lidavidm:
## What's Changed

There is no more need after
https://github.com/apache/arrow/commit/32641ecfd97e302ea8119b1147d77aced96d097b

### DIFF
--- a/.github/workflows/rc.yml
+++ b/.github/workflows/rc.yml
@@ -261,11 +261,6 @@ jobs:
           # bundled Protobuf.
           brew uninstall protobuf
 
-          # We need Flatbuffers 24, not the latest version
-          # Homebrew does not offer older versions, so remove the Homebrew
-          # package and rely on Arrow using a bundled version instead
-          brew uninstall flatbuffers
-
           brew bundle --file=Brewfile
       - name: Prepare ccache
         run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/rc.yml
```diff
@@ -281,6 +276,7 @@ jobs:
           set -e
           # make brew Java available to CMake
           export JAVA_HOME=$(brew --prefix openjdk@11)/libexec/openjdk.jdk/Contents/Home
+          export ARROW_BUILD_TESTS=OFF
```
lidavidm, line 279: I disabled this

(1) we're not running the tests in CI anyways, so we might as well save some time
(2) they fail to build currently due to whatever Boost/Homebrew just did

kou: How about disabling tests by default?

```diff
diff --git a/ci/scripts/jni_macos_build.sh b/ci/scripts/jni_macos_build.sh
index 65b255e2..f7543b6f 100755
--- a/ci/scripts/jni_macos_build.sh
+++ b/ci/scripts/jni_macos_build.sh
@@ -61,7 +61,7 @@ github_actions_group_begin "Building Arrow C++ libraries"
 install_dir="${build_dir}/cpp-install"
 : "${ARROW_ACERO:=ON}"
 export ARROW_ACERO
-: "${ARROW_BUILD_TESTS:=ON}"
+: "${ARROW_BUILD_TESTS:=OFF}"
 export ARROW_BUILD_TESTS
 : "${ARROW_DATASET:=ON}"
 export ARROW_DATASET
```

kou (approved):
+1

PR #718, opened 2025-04-14
GH-698: Improve and fix Avro read consumers

### LINKED ISSUE #698: Avro support - Improve existing read capabilities

Raising this ticket as a second step on Avro support, following on from #615.  On this ticket I'd like to cover:

1)   Round trip test cases for all supported data types (schema and data)
2)   Fix nullability handling - Avro union of [ null, type ] should be handled as a single nullable field / vector, not create an Arrow union
3)   Expose an API for creating an Arrow schema directly from an Avro schema with the same type mapping as the existing consumers
4)   Expose an API to allow a VSR to be recycled when reading data (the VSR should be resized to accommodate an Avro block)

Regarding point 2, I'm think of adding a flag to the AvroToArrowConfig class. By default the flag can be false to preserve the current behaviour.

I have started work on this ticket but will need to wait for #638 to merge before raising a draft PR.

I think there are two more PRs needed in this series, one to provide a high-level API  to read / write whole files, this would use the producers / consumers internally, understand Avro's block structure and map one full Arrow VSR to one Avro block. The last PR would be to add some extra features including compression and dictionary encoding / enums.

martin-traverse:
Hi @lidavidm - here is part 2 in my Avro series, apologies for the delay, it's the usual work / contention story!

## What's Changed

This PR relates to #698 and is the second in a series intended to provide full Avro read / write support in native Java. It adds round-trip tests for both schemas (Arrow schema -> Avro -> Arrow) and data (Arrow VSR -> Avro block -> Arrow VSR). It also adds a number of fixes and improvements to the Avro Consumers so that data arrives back in its original form after a round trip. The main changes are:

* Added a top level method in AvroToArrow to convert Avro schema directly to Arrow schema (this may exist elsewhere, but is needed to provide an API that matches the logic of this implementation)
* Avro unions of [ type, null ] or [ null, type ] now have special handling, these are interpreted as a single nullable type rather than a union. Setting legacyMode = false in the AvroToArrowConfig object is required to enable this behaviour, otherwise unions are interpreted literally. Unions with more than 2 elements are always interpreted literally (but, per #108, in practice Java's current Union implementation is probably not usable with Avro atm).
* Added support for new logical types (decimal 256, timestamp nano and 3 local timestamp types)
* Existing timestamp-mills and timestamp-micros times now interpreted as zone-aware (previously they were interpreted as local, but now the local timestamp types are interpreted as local - I think this is correct per the [Avro spec](https://avro.apache.org/docs/1.12.0/specification/#timestamps)). Requires setting legacyMode = false.
* Removed namespaces from generated Arrow field names in complex types. E.g. the Avro field myNamepsace.outerRecord.structField.intField should be called just "intField" inside the Arrow struct. This doesn't affect the skip field logic, which still works using the qualified names. This requires setting legacyMode = false.
* Remove unexpected metadata in generated Arrow fields (empty alias lists and attributes interpreted as part of the field schema). This requires setting legacyMode = false.
* Use the expected child vector names for Arrow LIST and MAP types when reading. For LIST, the default child vector is called "$data$" which is illegal in Avro, so the child field name is also changed to "item" in the producers. This requires setting legacyMode = false.

Breaking changes have been removed from this PR.

Per discussion below, all breaking changes are now behind a "legacyMode" flag in the AvroToArrowConfig object, which is enabled by default in all the original code paths.

Closes #698 .

This change is meant to allow for round trip of schemas and individual Avro data blocks (one Avro data block -> one VSR). File-level capabilities are not included. I have not included anything to recycle the VSR as part of the read API, this feels like it belongs with the file-level piece. Also I have not done anything specific for enums / dict encoding as of yet.


adapter/avro/src/test/java/org/apache/arrow/adapter/avro/RoundTripSchemaTest.java
lidavidm, line 1: Is there an opportunity to structure this as a parameterized test?

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/AvroToArrowUtils.java
```diff
@@ -277,11 +333,17 @@ private static Consumer createConsumer(
...
       case BYTES:
         if (logicalType instanceof LogicalTypes.Decimal) {
-          arrowType = createDecimalArrowType((LogicalTypes.Decimal) logicalType);
+          LogicalTypes.Decimal decimalType = (LogicalTypes.Decimal) logicalType;
+          arrowType = createDecimalArrowType(decimalType, schema);
           fieldType =
               new FieldType(nullable, arrowType, /* dictionary= */ null, getMetaData(schema));
           vector = createVector(consumerVector, fieldType, name, allocator);
-          consumer = new AvroDecimalConsumer.BytesDecimalConsumer((DecimalVector) vector);
+          if (decimalType.getPrecision() <= 38) {
```
lidavidm, line 341: Hmm, it's technically possible to have a decimal256 with a smaller precision though

lidavidm: I guess in that case it would round-trip to the smaller type?

lidavidm (commented):
I think in the interest of trying to keep semver, we should avoid breaking changes if possible. Any thoughts @jbonofre @laurentgo? Or we could just call the next release 19.0.0...

If it helps we could just have a single flag for "old" behavior?

martin-traverse:
> I think in the interest of trying to keep semver, we should avoid breaking changes if possible. Any thoughts @jbonofre @laurentgo? Or we could just call the next release 19.0.0...
> 
> If it helps we could just have a single flag for "old" behavior?

A major version bump feels a bit extreme for something as small as this! Let me try the single flag approach. There might be complications with the change for zone-aware vs local timestamps, because the types are different, but I think so long as the check happens before the types are decided it should be ok.

martin-traverse:
Ok, here is an update. I have put a flag "legacyMode" in the config object and used that to control all the places where the old logic is impacted. I have allowed decimal 256 to come through in legacy mode and also timestamp-nanos with the old semantics. The local-timestamp-xxx types do not come through in legacy mode, because timestamp-xxx is already treated as local. I have replaced the original code for the logical types test and those are all passing.

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/ArrowToAvroUtils.java
```diff
@@ -332,7 +332,17 @@ private static <T> T buildBaseTypeSchema(
 
       case List:
       case FixedSizeList:
-        return buildArraySchema(builder.array(), field, namespace);
+        // Arrow uses "$data$" as the field name for list items, that is not a valid Avro name
+        Field itemField = field.getChildren().get(0);
+        if (ListVector.DATA_VECTOR_NAME.equals(itemField.getName())) {
```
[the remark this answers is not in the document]
martin-traverse: Hm, I think for list / map types using the constant defined names for children makes sense, with "item" instead of "$data$" for list items. More generally, we could normalise illegal chars to "_" to match the [Avro name rules](https://avro.apache.org/docs/1.12.0/specification/#names). Per my understanding similar rules are already enforced in C++, but are not part of the Arrow spec or Java implementation.

Very happy to put the normalisation in, it's probably a more useful behaviour than throwing an error in the adapter. Would you like me to do it?

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/producers/AvroNullableProducer.java
```diff
@@ -21,8 +21,8 @@
 import org.apache.avro.io.Encoder;
 
 /**
- * Producer wrapper which producers nullable types to an avro encoder. Write the data to the
- * underlying {@link FieldVector}.
+ * Producer wrapper which producers nullable types to an avro encoder. Reed data from the underlying
```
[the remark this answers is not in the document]
martin-traverse: ;-)

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/AvroToArrowConfig.java
```diff
@@ -81,4 +114,8 @@ public DictionaryProvider.MapDictionaryProvider getProvider() {
   public Set<String> getSkipFieldNames() {
     return skipFieldNames;
   }
+
+  public boolean isHandleNullable() {
```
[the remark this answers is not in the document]
martin-traverse: This is now part of the legacyMode parameter

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/AvroToArrowUtils.java
martin-traverse: For FIXED decimals I am using the fixedSize to choose the decimal type. For BYTES decimals yes they would just come back as the smaller type if the precision fits.

I used FIXED as the default output for decimals in the producers, because it is closer to the Arrow representation, but on reflection Avro as a format is very focused on keeping data compact, maybe BYTES makes more sense. It seems to be the default choice in Avro. Do you think we should go with that?

martin-traverse:
I have removed the breaking changes text from the headline text but am not able to remove the label.

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/AvroToArrowUtils.java
lidavidm: I think FIXED is okay to allow round-trip by default even if it's not technically as compact.

lidavidm (approved):

lidavidm:
Looks like there are some lint errors to be fixed

martin-traverse:
> Looks like there are some lint errors to be fixed

Apologies - I have reapplied spotless, should be ok now!

PR #722, opened 2025-04-22
GH-721: Allow using 1GB+ data buffers in variable width vectors

### LINKED ISSUE #721: Allow using 1GB+ data buffers in variable width vectors

In `BaseVariableWidthVector` the reallocation logic of the data buffer is to always use the "next power of 2" size instead of the actual desired size (see [here](https://github.com/apache/arrow-java/blob/main/vector/src/main/java/org/apache/arrow/vector/BaseVariableWidthVector.java#L574)). The reallocation fails if the new size exceeds the `MAX_BUFFER_SIZE` which is `Integer.MAX_VALUE` by default.
The issue is `Integer.MAX_VALUE` is just 1 byte less than a power of 2 (2<sup>31</sup> - 1) which means that exceeding the desired capacity of `1073741824` (1GB), the reallocation would fail. It means, we are unable to get a data buffer larger than 1GB while the default maximum size is almost 2GB.

gszadovszky:
## What's Changed

Allow actually reaching MAX_BUFFER_SIZE at reallocating variable width vectors instead of exceeding it calculating the next power of 2.
For unit testing the maximum allocation size has be increased to 2MB - 1byte to simulate the default maximum behavior. Due to this change needed some updates in existing unit tests because of the round ups used at calculating the required buffer sizes.

Closes #721.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -327,8 +327,8 @@ under the License.
               <io.netty.tryReflectionSetAccessible>true</io.netty.tryReflectionSetAccessible>
               <user.timezone>UTC</user.timezone>
               <!-- Note: changing the below configuration might increase the max allocation size for a vector
-              which in turn can cause OOM. -->
-              <arrow.vector.max_allocation_bytes>1048576</arrow.vector.max_allocation_bytes>
+              which in turn can cause OOM. Using 1MB - 1byte to simulate the defaul limit of 2^31 - 1 bytes. -->
+              <arrow.vector.max_allocation_bytes>1048575</arrow.vector.max_allocation_bytes>
             </systemPropertyVariables>
             <useModulePath>false</useModulePath>
           </configuration>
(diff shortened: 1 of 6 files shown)


lidavidm (commented):
Is it possible to write a test for this?

gszadovszky:
> Is it possible to write a test for this?

I think we already have tests related to reallocation. The contract of the related methods did not change related to this change. Do you think we still need to test that we are allocating between 1GB and 2GB?

BTW, do you know what the actual failure is about "Ensure PR format"?

lidavidm:
You can ignore that check for now.

I ask because previously allocating between 1 and 2 GB didn't work, so a regression test would be useful to provide.

gszadovszky:
@lidavidm, I'm not sure how could I test this. The max allocation size is decreased for the unit tests so I cannot test reallocations for 1GB+ sizes. Increasing this setting would risk OOM issues in the CI. See [here](https://github.com/apache/arrow-java/blob/main/pom.xml#L329-L331).

lidavidm (approved):
Thanks!

gszadovszky:
Thank you, @lidavidm!

PR #741, opened 2025-05-09
MINOR: Bump version to 18.3.0

wgtmac:


jbonofre:
That's not correct. It should be SNAPSHOT version as the release is done by the maven-release-plugin.

kou (approved):
+1

This is correct.

We don't create a release branch. We release a RC from main. So we should use no `-SNAPSHOT` version here.

kou:
FYI: We don't use maven-release-plugin for our release.

kou:
> FYI: We don't use maven-release-plugin for our release.

Ah, sorry. We use maven-release-plugin. https://github.com/apache/arrow-java/blob/main/dev/release/bump_version.sh that creates this PR uses maven-release-plugin.

kou:
Do you still think we need `-SNAPSHOT` version in this PR?
(I think that we don't need `-SNAPSHOT` version. Our next release script https://github.com/apache/arrow-java/blob/main/dev/release/release_rc.sh just tags the commit that merges this PR.)

wgtmac:
I think it is fine. The caveat is that the main branch must be kept unchanged until the release is done.

kou:
> The caveat is that the main branch must be kept unchanged until the release is done.

Right. It's a restriction of the current release process.

lidavidm (approved):

wgtmac:
Are you good with this PR? @jbonofre

jbonofre (approved):
Ok as soon as the version is bumped to SNAPSHOT after the release

wgtmac:
Thanks! Let me merge it and proceed.

PR #754, opened 2025-05-13
MINOR: Bump version to 19.0.0-SNAPSHOT

wgtmac:


lidavidm:
Do we actually want the next release to be 19.0.0? Or if we aren't making any breaking changes we can do 18.4.0?

lidavidm:
I guess we can always change it at release time (we can change the milestone name to just be 'Next' or something?)

jbonofre:
I think it's ok to set 19.0.0-SNAPSHOT and bump the version to target release at release time.

jbonofre (approved):

wgtmac:
Yes, I was hesitant to choose from `18.4.0-SNAPSHOT` and `19.0.0-SNAPSHOT`. It can always be decided at the release time depending on the actual commits.

PR #760, opened 2025-05-20
fix get length of byte[] in TryCopyLastError

### LINKED ISSUE #759: coredump in `TryCopyLastError`

If `Consumer` throws exception, `TryCopyLastError` would copy exception message to C buffer. `TryCopyLastError` assume the string(byte[]) returned by `lastError` is '\0' terminated and test the length of the string by `strlen` which result in invalid memory access. In actually, `lastError` is a UTF8 encoded buffer(byte[]).

The code of `TryCopyLastError`:

```c
  jobject error_data =
      env->GetObjectField(private_data->j_private_data_, kPrivateDataLastErrorField);
  ...
  auto arr = reinterpret_cast<jbyteArray>(error_data);
  jbyte* error_bytes = env->GetByteArrayElements(arr, nullptr);
  if (!error_bytes) {
    private_data->last_error_.clear();
    return;
  }

  char* error_str = reinterpret_cast<char*>(error_bytes);
  private_data->last_error_ = std::string(error_str, std::strlen(error_str));
```

The code below shows how `lastError` be setted:
```java
private int setLastError(Throwable err) {
      // Do not let exceptions propagate up to JNI
      try {
        StringWriter buf = new StringWriter();
        PrintWriter writer = new PrintWriter(buf);
        err.printStackTrace(writer);
        lastError = buf.toString().getBytes(StandardCharsets.UTF_8);
      } catch (Throwable e) {
        // Bail out of setting the error message - we'll still return an error code
        lastError = null;
      }
      return 5; // = EIO
    }
```

hnwyllmm:
## What's Changed

We should get the length of byte[] by `GetArrayLength` but not `strlen` which may cause invalid memory access.

Closes #759 .

### DIFF
--- a/c/src/main/cpp/jni_wrapper.cc
+++ b/c/src/main/cpp/jni_wrapper.cc
@@ -205,8 +205,9 @@ void TryCopyLastError(JNIEnv* env, InnerPrivateData* private_data) {
     return;
   }
 
+  jsize error_bytes_len = env->GetArrayLength(arr);
   char* error_str = reinterpret_cast<char*>(error_bytes);
-  private_data->last_error_ = std::string(error_str, std::strlen(error_str));
+  private_data->last_error_ = std::string(error_str, error_bytes_len);
 
   env->ReleaseByteArrayElements(arr, error_bytes, JNI_ABORT);
 }


lidavidm (commented):
Is it possible to add a test?

hnwyllmm:
The error message in CI doesn't seem to be related to my PR:
```text
/opt/rh/devtoolset-10/root/usr/libexec/gcc/x86_64-redhat-linux/10/ld: cannot find -lutf8proc
  collect2: error: ld returned 1 exit status
```

linkage: https://github.com/apache/arrow-java/actions/runs/15140171347/job/42594091462?pr=760

hnwyllmm:
> Is it possible to add a test?

I'll try to add a test case in `integration_tests.py`.

kou:
> The error message in CI doesn't seem to be related to my PR:
> 
> ```
> /opt/rh/devtoolset-10/root/usr/libexec/gcc/x86_64-redhat-linux/10/ld: cannot find -lutf8proc
>   collect2: error: ld returned 1 exit status
> ```
> 
> linkage: https://github.com/apache/arrow-java/actions/runs/15140171347/job/42594091462?pr=760
> 
> Do you have some suggestions?

Could you open a separated issue for it?
Let's fix it as a separated task.

lidavidm:
> > Is it possible to add a test?
> 
> I'll try to add a test case in `integration_tests.py`.

Do you need an integration test? IMO, you should be able to export a stream, then immediately import it.

kou:
Thanks. (We've fixed this problem in apache/arrow. Sorry for not sharing it here...)

hnwyllmm:
> > > Is it possible to add a test?
> > 
> > 
> > I'll try to add a test case in `integration_tests.py`.
> 
> Do you need an integration test? IMO, you should be able to export a stream, then immediately import it, all within Java.

I add a unittest but I think it can't make sure `TryCopyLastError` works correctly, so I suggest it should be removed.
We can read memory after the error string address in most cases. We usually check such problems by sanity tools.

lidavidm:
Can't the test validate that the exception has the right error message after making it through the FFI boundary?

hnwyllmm:
> Can't the test validate that the exception has the right error message after making it through the FFI boundary?

PTAL.
I use a real exception message in the ExceptionTest.

c/src/test/java/org/apache/arrow/c/ExceptionTest.java
```diff
@@ -0,0 +1,172 @@
...
+
+      root.allocateNew();
+      ints.setSafe(0, 1);
+      ints.setSafe(1, 2);
+      ints.setSafe(2, 4);
+      ints.setSafe(3, 8);
+      root.setRowCount(4);
+      batches.add(unloader.getRecordBatch());
+
+      final String exceptionMessage = "java.lang.RuntimeException: Error occurred while getting next schema root.\n\tat org.apache.arrow.adapter.jdbc.ArrowVectorIterator.next(ArrowVectorIterator.java:205)\n\tat com.oceanbase.external.jdbc.JdbcScanner.loadNextBatch(JdbcScanner.java:73)\n\tat org.apache.arrow.c.ArrayStreamExporter$ExportedArrayStreamPrivateData.getNext(ArrayStreamExporter.java:72)\nCaused by: java.lang.RuntimeException: Error occurred while consuming data.\n\tat org.apache.arrow.adapter.jdbc.ArrowVectorIterator.consumeData(ArrowVectorIterator.java:127)\n\tat org.apache.arrow.adapter.jdbc.ArrowVectorIterator.load(ArrowVectorIterator.java:178)\n\tat org.apache.arrow.adapter.jdbc.ArrowVectorIterator.next(ArrowVectorIterator.java:198)\n\t... 2 more\nCaused by: java.lang.OutOfMemoryError: Java heap space\n";
```
lidavidm, line 64: Um, this is a little excessive. Can we just use a short canary string?

c/src/test/java/org/apache/arrow/c/ExceptionTest.java
```diff
@@ -0,0 +1,172 @@
...
+              }
+              continue;
+            }
+            loader.load((ArrowRecordBatch) batch);
+
+            assertThat(reader.getVectorSchemaRoot().getRowCount()).isEqualTo(root.getRowCount());
+          }
+        }
+      }
+    }
```
lidavidm, line 98: We need to assert that we did indeed get an exception.

c/src/test/java/org/apache/arrow/c/ExceptionTest.java
```diff
@@ -0,0 +1,172 @@
...
+import org.apache.arrow.vector.ipc.ArrowReader;
+import org.apache.arrow.vector.ipc.message.ArrowRecordBatch;
+import org.apache.arrow.vector.types.pojo.ArrowType;
+import org.apache.arrow.vector.types.pojo.Field;
+import org.apache.arrow.vector.types.pojo.Schema;
+import org.junit.jupiter.api.Test;
+
+final class ExceptionTest {
+  @Test
+  public void testException() throws IOException {
```
lidavidm, line 46: Please indicate somewhere that this is a regression test for #759.

c/src/test/java/org/apache/arrow/c/ExceptionTest.java
```diff
@@ -0,0 +1,184 @@
...
+          for (Object batch : batches) {
+            try {
+              reader.loadNextBatch();
+            } catch (Exception e) {
+              assertThat(exceptionThrowed).isEqualTo(null);
+              final String eMessage = e.getMessage();
+              // 1 for '}', ref to CDataJniException
+              assertThat(eMessage.length()).isGreaterThan(expectExceptionMessage.length() + 1); 
+              assertThat(eMessage.substring(eMessage.length() - expectExceptionMessage.length() - 1, eMessage.length() - 1))
+                  .isEqualTo(expectExceptionMessage);
```
lidavidm, line 99: Just use `contains`...

c/src/test/java/org/apache/arrow/c/ExceptionTest.java
```diff
@@ -0,0 +1,184 @@
...
+              exceptionThrowed = e;
+              continue;
+            }
```
lidavidm, line 102: Use assertThrows.

lidavidm (changes requested):

c/src/test/java/org/apache/arrow/c/ExceptionTest.java
hnwyllmm: No, we must make sure the string in the exception is the same with we throws.
"abc\x037" contains "abc", but it's not correct and we want to catch the bug.

hnwyllmm: No.
Here is the CDataJniException::getMessage:
```java
public String getMessage() {
        return "CDataJniException{errno=" + this.errno + ", message=" + super.getMessage() + '}';
    }
```
The message of the exception we thrown is `super().getMessage()`.

lidavidm (approved):

lidavidm:
ah right, CI is broken...

hnwyllmm:
Thanks for your patient. It seems the failed CI test cases are not related with this PR now. Please take a look.

lidavidm:
Ok. I'll merge this, but the CI really needs to be fixed...I just don't have time for arrow-java these days :/

PR #764, opened 2025-05-23
MINOR: Fix broken Linux JNI build

lidavidm:
## What's Changed

I think this was due to apache/arrow@1229ceddd512aa5b3a601019ee3b4bae28cfa3e6.

We need to explicitly set ARROW_VCPKG=ON for Linux builds so that Findutf8proc.cmake does the right thing.

### DIFF
--- a/ci/scripts/jni_build.sh
+++ b/ci/scripts/jni_build.sh
@@ -50,11 +50,15 @@ Darwin)
 esac
 
 : "${ARROW_JAVA_BUILD_TESTS:=${ARROW_BUILD_TESTS:-ON}}"
+: "${ARROW_VCPKG:=OFF}"
 : "${CMAKE_BUILD_TYPE:=release}"
 read -ra EXTRA_CMAKE_OPTIONS <<<"${JAVA_JNI_CMAKE_ARGS:-}"
+# Must set ARROW_VCPKG because upstream Findutf8proc.cmake checks that to
+# determine what to do
 cmake \
   -S "${source_dir}" \
   -B "${build_dir}" \
+  -DARROW_VCPKG="${ARROW_VCPKG}" \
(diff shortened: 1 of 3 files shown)


ci/scripts/jni_build.sh
```diff
@@ -50,11 +50,15 @@ Darwin)
...
 : "${ARROW_JAVA_BUILD_TESTS:=${ARROW_BUILD_TESTS:-ON}}"
+: "${ARROW_VCPKG:=OFF}"
 : "${CMAKE_BUILD_TYPE:=release}"
 read -ra EXTRA_CMAKE_OPTIONS <<<"${JAVA_JNI_CMAKE_ARGS:-}"
+# Must set ARROW_VCPKG because upstream Findutf8proc.cmake checks that to
+# determine what to do
 cmake \
   -S "${source_dir}" \
   -B "${build_dir}" \
+  -DARROW_VCPKG="${ARROW_VCPKG}" \
```
kou, line 61: Can we use `ARROW_DEPENDENCY_SOURCE=VCPKG` instead? It sets `ARROW_VCPKG` internally.

lidavidm: I tried this and it doesn't seem to work: I don't think the Java JNI CMake code has anything to detect this. Also, `Findutf8proc.cmake` installed by Arrow's CMake checks `ARROW_VCPKG`, but I don't see any other code installed to set this.

kou: Ah, OK. I'll fix it in apache/arrow.

kou: https://github.com/apache/arrow/pull/46565 will fix this.

lidavidm:
After https://github.com/apache/arrow/issues/46564 I re-ran the job on main and it passes now, so I'm closing this. Thanks Kou!

Issue #763, opened 2025-05-22
Data import methods free memory


pepijnve:
The various import methods in the Data class are documented to transfer ownership of the Arrow data from the input object to internal values as per the C FFI. If I'm reading https://arrow.apache.org/docs/format/CDataInterface.html#moving-an-array correctly, implementations are supposed to hollow out the input object in that case by nulling the release field. In practice a class like ArrayImporter does this

```
void importArray(ArrowArray src) {
  ...
  // Move imported array
  ArrowArray ownedArray = ArrowArray.allocateNew(allocator);
  ownedArray.save(snapshot);
  src.markReleased();
  src.close();
  ...
}
```

So not only is the array marked as released, it is also closed so the ArrowArray instance is no longer usable. This may or may not happen when you call `Data#importIntoVector` for instance, since that might abort early when given a null `Allocator` which makes the post conditions of `Data#importIntoVector` a bit unclear. You kind of expect the ArrowArray to have been closed, but you still need to call close yourself as well or risk having a memory leak.

What's the rationale behind the choice to also call `close` on objects passed to Data rather than expecting people to use try-with-resources/try-finally themselves?

lidavidm:
Basically, the idea is that the importer should not free the `struct ArrowArray` itself, so that the caller code can use its own try-with-resources? I think that makes sense. (Well, either that, or the importer internally should free the struct even if it errors, which I think is less clean but would be less of a breaking change.)

pepijnve:
> Basically, the idea is that the importer should not free the struct ArrowArray itself, so that the caller code can use its own try-with-resources?

I think that would make more sense personally, but it would be a serious backwards incompatible change. The alternative would be to add additional methods (perhaps an overloaded method signature with a boolean parameter, although that's not really pretty either) that let the caller control what happens.

I would be happy to put together a PR for this if you like. Any preference on what this should look like exactly?

lidavidm:
New methods is probably cleanest

PR #766, opened 2025-05-23
GH-765: Do not close/free imported BaseStruct objects

### LINKED ISSUE #765: Avoid closing/freeing struct-like object instances on import

Currently the various import methods in `org.apache.arrow.c.Data` call `close` on the `BaseStruct` arguments they receive as parameter. This prevents callers from reusing these objects multiple times. Since the caller needs to allocate these objects themselves, it seems like it would be more consistent to also expect the caller to free them. This works well in combination with try-with-resources.

See related discussion #763.

pepijnve:
## What's Changed

This PR removes the direct and indirect calls to `BaseStruct#close` from `org.apache.arrow.c.Data`. By not eagerly closing/freeing these objects callers can reuse instances multiple times.

**This contains breaking changes.**: a quick check in, for instance, `org.apache.arrow.c.StreamTest` shows that a lot of existing code is already written using try-with-resources. This change will not have any impact there. Code that does not use a try-with-resources or try-finally-close pattern and instead counts on `Data` to call `close` will report memory leaks after this change.

Closes #765.


pepijnve:
https://github.com/apache/arrow-java/issues/763#issuecomment-2903439789 suggests adding new or overloaded methods that do not call close instead of introducing a breaking change. Is there any preference wrt naming these?

lidavidm:
Perhaps by analogy with the base library, `load` and `unload`, instead of `import` and `export`? That won't tell you which method is newer, but then we can add `Deprecated` annotations

pepijnve:
Just FYI, I'm experimenting a bit, and neither the `load` variant nor the boolean flag variant are particularly appealing.

`load` feels a bit asymmetric. There's no need for any changes in the `export` methods as far as I can tell, so then you have three sets of methods: `export`, `import`, and `load`.

The overload of `import` with a boolean flag is an option, but a bit ugly for `importVectorSchemaRoot(BufferAllocator,      ArrowArray, ArrowSchema, CDataDictionaryProvider)` where the behavior is controlled for both the array and the schema. Having two boolean flags seems even worser though.

pepijnve:
I've pushed a second attempt that introduces overloaded import methods with a boolean argument. I've moved all the close calls to the Data class for clarity. `ArrayImporter` and `ArrowArrayStreamReader` are both package visible classes, so the change there is not a breaking change.

lidavidm:
checkstyle is still unhappy:

```
Warning:  src/main/java/org/apache/arrow/c/Data.java:[361] (javadoc) SummaryJavadoc: First sentence of Javadoc is missing an ending period.
Warning:  src/main/java/org/apache/arrow/c/Data.java:[543] (javadoc) AtclauseOrder: Javadoc comment at column 36 has parse error. Details: mismatched input '#' expecting MEMBER while parsing REFERENCE
Warning:  src/main/java/org/apache/arrow/c/Data.java:[543] (javadoc) JavadocTagContinuationIndentation: Javadoc comment at column 36 has parse error. Details: mismatched input '#' expecting MEMBER while parsing REFERENCE
Warning:  src/main/java/org/apache/arrow/c/Data.java:[543] (javadoc) NonEmptyAtclauseDescription: Javadoc comment at column 36 has parse error. Details: mismatched input '#' expecting MEMBER while parsing REFERENCE
Warning:  src/main/java/org/apache/arrow/c/Data.java:[543] (javadoc) SummaryJavadoc: Javadoc comment at column 36 has parse error. Details: mismatched input '#' expecting MEMBER while parsing REFERENCE
```

pepijnve:
I've added a couple of test. `mvn verify` says :+1: locally.

lidavidm (approved):

Issue #70, opened 2024-06-25
[Java] Usage of Mix of assertj in favor of hamcrest 


vibhatha:
### Describe the enhancement requested

Through out the unit tests, we have a mix usage of assertj and hamcrest and specifically in dataset and flight modules, hamcrest is being widely used. That must be replaced with assertj.  

### Component(s)

Java

PR #772, opened 2025-05-26
GH-70: Move from `hamcrest` to `assertj` in `flight-sql`

rtadepalli:
## What's Changed
Series of PRs to consolidate on using `assertj` in tests as part of https://github.com/apache/arrow-java/issues/70.

### DIFF
--- a/flight/flight-sql/src/test/java/org/apache/arrow/flight/sql/test/TestFlightSql.java
+++ b/flight/flight-sql/src/test/java/org/apache/arrow/flight/sql/test/TestFlightSql.java
@@ -23,7 +23,6 @@ import static org.apache.arrow.flight.sql.util.FlightStreamUtils.getResults;
 import static org.apache.arrow.util.AutoCloseables.close;
 import static org.hamcrest.CoreMatchers.containsString;
 import static org.hamcrest.CoreMatchers.is;
-import static org.hamcrest.CoreMatchers.notNullValue;
 import static org.hamcrest.CoreMatchers.nullValue;
 import static org.junit.jupiter.api.Assertions.assertAll;
 import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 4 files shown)


flight/flight-sql/src/test/java/org/apache/arrow/flight/sql/test/TestFlightSql.java
```diff
@@ -915,28 +914,28 @@ public void testGetCommandExportedKeys() throws Exception {
...
-              nullValue(String.class), // fk_catalog_name
-              is("APP"), // fk_schema_name
-              is("INTTABLE"), // fk_table_name
-              is("FOREIGNID"), // fk_column_name
-              is("1"), // key_sequence
-              containsString("SQL"), // fk_key_name
-              containsString("SQL"), // pk_key_name
-              is("3"), // update_rule
-              is("3")); // delete_rule
+              new HamcrestCondition<>(nullValue(String.class)), // pk_catalog_name
```
rtadepalli, line 919: Not particularly certain how to get rid of Hamcrest entirely here. I was aware of `HamcrestCondition` so using that, but open to suggestions if anyone knows how this list-based matching can be accomplished solely using assertj.

rtadepalli:
I've formatted my changes using `mvn spotless:apply`.

flight/flight-sql/src/test/java/org/apache/arrow/flight/sql/test/TestFlightSql.java
```diff
@@ -247,16 +247,15 @@ private static List<List<String>> getNonConformingResultsForGetSqlInfo(
   @Test
   public void testGetTablesSchema() {
     final FlightInfo info = sqlClient.getTables(null, null, null, null, true);
-    MatcherAssert.assertThat(
-        info.getSchemaOptional(), is(Optional.of(FlightSqlProducer.Schemas.GET_TABLES_SCHEMA)));
+    Assertions.assertThat(info.getSchemaOptional())
```
lidavidm, line 250: nit, but IMO it's OK to statically import `assertThat`.

flight/flight-sql/src/test/java/org/apache/arrow/flight/sql/test/TestFlightSql.java
lidavidm: Using a list of lambdas?

lidavidm (commented):
Thanks for doing this!

PR #775, opened 2025-05-29
GH-774: Consoliate `BitVectorHelper.getValidityBufferSize` and `BaseValueVector.getValidityBufferSizeFromCount`

### LINKED ISSUE #774: Remove duplicate function `getValidityBufferSizeFromCount` in `BaseValueVector`

This is a duplicated function of `BitVectorHelper.getValidityBufferSize`. Unify these two without letting multiple functions that call the same underlying function linger around.

rtadepalli:
## What's Changed
Just removing some duplicate functions in anticipation of cleaning out some transferPair duplication across complex vectors.

Closes #774

### DIFF
--- a/vector/src/main/java/org/apache/arrow/vector/BaseValueVector.java
+++ b/vector/src/main/java/org/apache/arrow/vector/BaseValueVector.java
@@ -110,11 +110,6 @@ public abstract class BaseValueVector implements ValueVector {
     return buffer;
   }
 
-  /* number of bytes for the validity buffer for the given valueCount */
-  protected static int getValidityBufferSizeFromCount(final int valueCount) {
-    return DataSizeRoundingUtil.divideBy8Ceil(valueCount);
-  }
-
   /* round up bytes for the validity buffer for the given valueCount */
   private static long roundUp8ForValidityBuffer(long valueCount) {
     return ((valueCount + 63) >> 6) << 3;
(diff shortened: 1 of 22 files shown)


lidavidm (commented):
Hmm. This is a protected API, so this is technically a breaking change. Albeit not one likely to be noticed. But I think we should mark it as such all the same, and perhaps consider making the next version 19.0.0 then. (Any thoughts from other reviewers?) The API has also been there for quite some time, though I suppose it's fine to clean things up when we notice redundancy.

lidavidm:
A protected method is still visible if a user defines their own subclass.

rtadepalli:
I guess nothing is being "removed" from the code base. If you / other reviewers have seen anything either on the mailing list or via GH issues about people implementing custom types then maybe marking this a breaking change is the way to go. If not then IMHO the change is low-key and likely won't impact anything. In the off chance someone is using custom types and they are impacted because of this, maybe we could do a minor release instead of a patch (don't know if that is any better).

vector/src/main/java/org/apache/arrow/vector/BaseValueVector.java
```diff
@@ -110,7 +110,14 @@ protected ArrowBuf releaseBuffer(ArrowBuf buffer) {
...
 
-  /* number of bytes for the validity buffer for the given valueCount */
+  /**
+   * Compute the size of validity buffer required to manage a given number of elements in a vector.
+   *
+   * @param valueCount number of elements in the vector
+   * @return buffer size
+   * @deprecated -- use {@link BitVectorHelper#getValidityBufferSizeFromCount} instead.
+   */
+  @Deprecated
```
lidavidm, line 120:
```suggestion
  @Deprecated(forRemoval = true, since = "18.4.0")
```

lidavidm (approved):

lidavidm:
CI is failing; there are still uses of the 'old' method around it seems?

rtadepalli:
Fixed now. Re-adding that method overrode my static import, have to refer to `BitVectorHelper` explicitly now.

lidavidm (approved):

PR #783, opened 2025-06-05
GH-155: Delegate opening mTLS cert files to `build()`

### LINKED ISSUE #155: [Java][FlightRPC] FlightServer builder should store File, not InputStream

Methods like useTls and useMTlsClientVerification should store a File instead of opening the file immediately, then at build time we can open the file and make sure to close the input stream (as required by Netty). This minimizes the logic in the builder setters. Followup for apache/arrow#38461

### Component(s)

FlightRPC, Java

rtadepalli:
## What's Changed
Make `FlightServer.Builder` store `File` and delegate opening the file to the actual `build()` method.

Closes #155.

### DIFF
--- a/flight/flight-core/src/main/java/org/apache/arrow/flight/FlightServer.java
+++ b/flight/flight-core/src/main/java/org/apache/arrow/flight/FlightServer.java
@@ -572,17 +572,17 @@ public class FlightServer implements AutoCloseable {
     }
 
     private void prepareTlsSettings() throws IOException {
-      if (this.keyFile != null) {
-        closeKey();
-        this.key = new FileInputStream(this.keyFile);
+      if (this.mTlsCACertFile != null) {
+        closeMTlsCACert();
+        this.mTlsCACert = new FileInputStream(this.mTlsCACertFile);
       }
       if (this.certChainFile != null) {
         closeCertChain();
         this.certChain = new FileInputStream(this.certChainFile);
       }
-      if (this.mTlsCACertFile != null) {
-        closeMTlsCACert();
-        this.mTlsCACert = new FileInputStream(this.mTlsCACertFile);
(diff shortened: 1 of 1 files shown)


rtadepalli:
Could I please get a review @lidavidm thanks.

lidavidm:
Sorry - I have basically 0 time for arrow-java recently. I will try to get to all the reviews soon but I can't make any promises.

flight/flight-core/src/main/java/org/apache/arrow/flight/FlightServer.java
```diff
@@ -552,5 +570,20 @@ public Builder producer(FlightProducer producer) {
       this.producer = Preconditions.checkNotNull(producer);
       return this;
     }
+
+    private void prepareTlsSettings() throws IOException {
```
lidavidm, line 574: The way this is written, if you set the File version then the deprecated version never takes effect, even if you set it afterwards. I suppose that's OK given the deprecation.

flight/flight-core/src/main/java/org/apache/arrow/flight/FlightServer.java
```diff
@@ -213,6 +222,16 @@ public static final class Builder {
 
     /** Create the server for this builder. */
     public FlightServer build() {
+      // Get TLS info in order if the server is being configured to use mTLS.
+      try {
+        prepareTlsSettings();
+      } catch (IOException e) {
+        closeMTlsCACert();
+        closeCertChain();
+        closeKey();
```
lidavidm, line 231: Honestly, this should be in a try-finally wrapping the entire build() call (since other exceptions in between this and the actual TLS settings will leak the handles). Either that, or this should be moved down to right before we use the file handles (and should still be in a try-finally)

PR #802, opened 2025-07-21
GH-794: Avro file read and write support

### LINKED ISSUE #794: Avro adapter - Read and write Avro container files

Part 4 in the Avro series, following on from #731. This will allow reading and writing whole files in the Avro container format as a series of batches. Each batch will correspond to one Avro file block and fill a single VSR. The VSR can be recycled between batches. Input and output can be to Avro encoder / decoder (set up externally) or to Java's native byte channels (which are set up with default binary encoder / decoder). To cater for async scenarios, the reader API should know how many bytes are required for a block before attempting to read it.

I'd like to propose the following API - hopefully this is going in the right direction. I've taken some inspiration from ArrowFilleReader / Writer and Json Reader / Writer, but it's not identical (and they're not identical to each other). If there is a desire to line up on specific naming / conventions then certainly happy to do that, in which case I'll need a steer on exactly how it should be. Otherwise if anyone has radically different ideas of what it should look like, please do share!

    class AvroFileWriter {

        // Writer owns a channel / encoder and will close them
        // VSR and optional dictionaries are not owned and will not be closed
        // VSR can be recycled or supplied as a stream

        // Avro encoder configured externally
        public AvroFileWriter(
            Encoder encoder,
            VectorSchemaRoot firstBatch, 
            DictionaryProvider dictionaries)

        // Sets up a defaulr binary encoder for the channel
        public AvroFileWriter(
            WritableByteChannel channel,
            VectorSchemaRoot firstBatch, 
            DictionaryProvider dictionaries)

        // Write the Avro header (throws if already written)
        void writeHeader()

        // Write the contents of the VSR as an Avro data block
        // Writes header if not yet written
        // Expects new data to be in the batch (i.e. VSR can be recycled)
        void writeBatch()

        // Reset vectors in all the producders
        // Supports a stream of VSRs if source VSR is not recycled
        void resetBatch(VectorSchemaRoot batch)

        // Closes encoder and / or channel
        // Does not close VSR or dictionary vectors
        void close()

    }
  
Now writing data looks like this:

    void writeAvro(MyApp app) {

        var root = app.prepareVsr();
        var dictionaries = app.prepareDictionaries()

        try (var writer = new AvroFileWriter(app.openChannel(), root, dictionaries)) {

            writer.writeHeader();

            // Assume recycling, loadBatch() puts fresh data into root
            while (app.loadBatch()) {
                writer.writeBatch()
            }
        }
    }

And then for the reader:

    class AvroFileReader implements DictionaryProvider {

        // Writer owns a channel / decoder and will close them
        // Schema / VSR / dictionaries are created when header is read
        // VSR / dictionaries are cleaned up on close
        // Dictionaries accessible through DictionaryProvider iface

        // Avro decoder configured externally
        public AvroFileWriter(
            Decoder decoder,
            BufferAllocator allocator)

        // Sets up a defaulr binary deocder for the channel
        // Avro read sequentially so seekable channel not needed
        public AvroFileWriter(
            ReadableByteChannel channel,
            BufferAllocator allocator)

        // Read the Avro header and set up schema / VSR / dictionaries
        void readHeader()

        // Schema and VSR available after readHeader() 
        Schema getSchema()
        VectorSchemaRoot getVectorSchemaRoot()

        // Read the next Avro block and load it into the VSR
        // Return true if successful, false if EOS
        // Also false in non-blocking mode if need more data
        boolean readBatch()

        // Check for position and size of the next Avro data block
        // Provides a mechanism for non-blocking / reactive styles
        boolean hasNextBatch();
        long nextBatchPosition();
        long nextBatchSize();

        // Closes encoder and / or channel
        // Also closes VSR and dictionary vectors
        void close()

    }

So reading looks like this:

    // Blocking style
    void readAvro(MyApp app) {

        try (var reader = new AvroFileReader(app.openChannel(), app.allocator()) {

            reader.readHeader();
            
            app.setSchema(reader.getSchema());
            app.setVsr(reader.getVectorSchemaRoot());
            app.setDictionaries(reader);

            while (reader.readBatch())) {
                app.saveBatch();
            }
        }
    }

    // Non-blocking stage to process one batch
    CompletionStage<Boolean> readAvroAsync(AvroFileReader reader) {

        if (reader.hasNextBatch()) {

            var start = reader.nextBatchStart();
            var end = reader.nextBatchEnd();

            return app.ensureBytesAvailable(start, end)
                .thenApply(x -> {

                    if (reader.readBatch()) {
                        app.saveBatch();
                    }

                    return reader;
                })
                .thenCompose(this::readAsync);
        }
        else {
            return CompletableFuture.completedFuture(true);
        }
    }

The non-blocking read is quite important for me as I have a web service that receives bytes in a stream. There is a slight gotcha because we need the first 8 bytes of the next batch before we know its size, but we can implement hasNextBatch() without them and the probably expose the batch padding size as a constant. 

Compression is probably worth thinking about now - each block is compressed individually so the implementation needs to treat the contents of each block as a separate chunk, that can be fed through a codec. My guess is this is fairly straightforward for codecs that are already available so we might as well include it rather than reworking later.

If this looks broadly right I'll make a start on top of #779

martin-traverse:
## What's Changed

Add top level reader / writer classes for Avro files. This is a draft for discussion, I haven't written tests, doc comments etc. so will do all that once we agree the shape of the implementation is correct.

I have built a very simple implementation which just uses Avro's public components with the existing producers / consumers, following the Avro container file spec. Internally Avro uses input / output streams and heap allocated byte arrays, so I have based our reader / writer on those elements. I have recycled buffers / streams wherever it is possible, without breaking into Avro's internal structures.

For compression I am using the Avro own codec implementations. Instantiation via CodecFactory is restricted to Avro's own file handling package, but the codecs themselves are mostly public, with the exception of Snappy for some unknown reason. I'd be happy to raise a ticket and ask about that, or we could just copy the Snappy implementation in our own namespace (it is a simple wrapper on Xerial).

I did look at an alternative approach, using  ArrowBuf for the batch buffers with Arrow's codec implementations, which we could add to. However there were a couple of issues:

1. Since Avro uses streams / byte arrays internally, pretty much every way of getting to ArrowBuf involved going to a byte array first and then copying. To break out of that we'd need to reimplement large parts of Avro's file package, including encoders / decoders and shade some key classes in the Avro namespace.

2. Arrow's codec API assumes that compressed data is always written with the uncompressed size stored at the start of the output, which makes them unusable for other formats that don't do that, including Avro. We'd need to add a new API and implement the codecs again to handle resizing the output buffer.

Given these considerations, I  eventually came to think the very simple approach I've drafted might be the best option.  If there are performance benefits to be had by switching to ArrowBuf, Channels etc. we'd need to write a lot more code, which needs to stay in sync and be maintained etc. We can still add overload constructors to wrap channels for IO.

Final point on non-blocking mode for the reader. The approach I have used is just to insist that when blocking = false the input stream must support mark / reset, and then peak at the beginning of each batch to determine its size. Also, non-blocking readers are direct, which disable's Avro's internal buffering. I'm assuming that anyone using non-blocking will implement their own stream and buffering logic, in which case they can add mark / reset and won't want Avro randomly reading extra bytes for an internal buffer! At least that is what I'm planning to do. I haven't tried to estimate header size - there isn't really a way to do that without reading the whole header. We could have something  like headerBytesNeeded() to incrementally return the number of bytes still needed. The alternative is just to provide a couple of MB and assume it's enough, which probably works, but on reflection  I do think we should probably add this, just so the API for non-blocking is "complete".

Hope this all makes sense - please let me know your thoughts when you get a chance and then I'll do the last bit of work to get this ready.

Closes #794.


adapter/avro/src/main/java/org/apache/arrow/adapter/avro/AvroFileReader.java
```diff
@@ -0,0 +1,374 @@
...
+    byte[] magic = new byte[AVRO_MAGIC.length];
+    decoder.readFixed(magic);
+    headerSize += magic.length;
+
+    // Validate Avro magic
+    int validateMagic =
+        BinaryData.compareBytes(AVRO_MAGIC, 0, AVRO_MAGIC.length, magic, 0, AVRO_MAGIC.length);
+
+    if (validateMagic != 0) {
+      throw new RuntimeException("Invalid AVRO data file: The file is not an Avro file");
```
lidavidm, line 109: Is there a more specific exception we could throw? Or at least, an IOException?

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/AvroFileReader.java
```diff
@@ -0,0 +1,374 @@
...
+
+    org.apache.avro.Schema.Parser parser = new org.apache.avro.Schema.Parser();
+
+    try (InputStream schemaStream = new ByteArrayInputStream(buffer.array())) {
+      return parser.parse(schemaStream);
+    }
+  }
+
+  private String processCodec(ByteBuffer buffer) {
+
```
lidavidm, line 180: nit but why do many methods start with an extra blank line 😅

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/AvroFileReader.java
```diff
@@ -0,0 +1,374 @@
...
+    if (buffer != null && buffer.remaining() > 0) {
+      return new String(buffer.array(), StandardCharsets.UTF_8);
+    } else {
+      return DataFileConstants.NULL_CODEC;
+    }
+  }
+
+  // Schema and VSR available after readHeader()
+  Schema getSchema() {
+    if (avroSchema == null) {
```
lidavidm, line 190: Does it make sense to null-check the field itself (arrowSchema, not avroSchema)?

Should we use a Preconditions or Objects method?

lidavidm: Or, given that this check repeats a lot - have a `checkStarted` method?

adapter/avro/src/main/java/org/apache/arrow/adapter/avro/AvroFileReader.java
```diff
@@ -0,0 +1,374 @@
...
+            + batchBuffer.remaining()
+            + SYNC_MARKER_SIZE;
+
+    // Validate sync marker - mismatch indicates a corrupt file
+    int validateMarker =
+        BinaryData.compareBytes(
+            syncMarker, 0, SYNC_MARKER_SIZE, batchSyncMarker, 0, SYNC_MARKER_SIZE);
+
+    if (validateMarker != 0) {
+      throw new RuntimeException("Invalid AVRO data file: The file is corrupted");
```
lidavidm, line 249: ditto, can we throw a more specific exception? Maybe create a subclass of IOException to indicate invalid files?

lidavidm (commented):
Sorry for the long delay. Overall this looks reasonable to me.

PR #806, opened 2025-07-31
GH-804: Prepend JDBC FlightSQL version to user agent

### LINKED ISSUE #804: Specify driver version in user-agent

Currently, the server can't differentiate if a request was made using a raw Flight client or the JDBC Flight SQL driver. In both scenarios the user-agent is something like:

`user-agent: grpc-java-netty/1.73.0`

I'd like to suggest in the scenario of JDBC Flight SQL driver to append to the user-agent:
`JDBC Flight SQL Client <version>`

This is similar to what is already done in with ADBC GO:
`ADBC Flight SQL Driver (unknown or development build) grpc-go/1.72.0`

xborder:
## What's Changed

* Driver version is passed on to NettyBuilderClient to append it to the user-agent header
* NettyBuilderClient now prepends `JDBC Flight SQL Client <version>` to the header (e.g. `JDBC Flight SQL Client 19.0.0-SNAPSHOT grpc-java-netty/1.73.0`)

Closes #804.

### DIFF
--- a/flight/flight-core/src/main/java/org/apache/arrow/flight/grpc/NettyClientBuilder.java
+++ b/flight/flight-core/src/main/java/org/apache/arrow/flight/grpc/NettyClientBuilder.java
@@ -60,7 +60,6 @@ public class NettyClientBuilder {
   protected String overrideHostname = null;
   protected List<FlightClientMiddleware.Factory> middleware = new ArrayList<>();
   protected boolean verifyServer = true;
-  protected String userAgent;
 
   public NettyClientBuilder() {}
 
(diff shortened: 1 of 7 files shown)


flight/flight-core/src/main/java/org/apache/arrow/flight/grpc/NettyClientBuilder.java
```diff
@@ -130,6 +131,11 @@ public NettyClientBuilder verifyServer(boolean verifyServer) {
     return this;
   }
 
+  public NettyClientBuilder userAgent(String userAgent) {
```
lidavidm, line 134: I'm against adding to this interface - the intent is that for any further customization, you should directly use the returned NettyChannelBuilder

lidavidm (approved):

PR #811, opened 2025-08-06
GH-797: [JDBC] Fix PreparedStatement#execute for DML/DDL

### LINKED ISSUE #797: PreparedStatement.execute() returns True for DML/DDL queries

In Apache Arrow Flight SQL Prepared Statement execution, result set generating queries should be executed via `CommandPreparedStatementQuery` and queries that return only the number of affected rows should be executed via `CommandPreparedStatementUpdate`

In JDBC, `PreparedStatement.execute()` allows for running generic queries, including SELECT, DML and DDL. For DML/DDL, this method should return `False` and provide an update count, for SELECT it should return `True` and result set should be provided. 

The cause of the issue is that the driver always follows the Arrow Flight SQL communications path for result set generating queries, meaning, the query is always executed via `CommandPreparedStatementQuery` and never via `CommandPreparedStatementUpdate` This means that the driver tries to execute DML/DDL as if they were SELECT statements, which can lead to incorrect return values of PreparedStatement.execute()

Code snippet to reproduce:

```  
// connection is a previously established Connection instance to a Flight SQL server
String insertPreparedStatementSingleRow = "INSERT INTO my_table (name) VALUES (?)";
try (PreparedStatement pstmt = connection.prepareStatement(insertPreparedStatementSingleRow)) {
  pstmt.setString(1, "My name");
  boolean hasResultSet = pstmt.execute();
  if (hasResultSet) {
    System.out.println("Unexpected: INSERT returned a ResultSet");
    try (ResultSet rs = pstmt.getResultSet()) {
      // Handle unexpected ResultSet
      System.out.println("Unexpected ResultSet...");
    }
  } else {
    // false = Update count available
    int updateCount = pstmt.getUpdateCount();
    System.out.println("Update count: " + updateCount + " row(s) affected");
  }
}
```

This was tested with a Dremio Arrow Flight SQL server, but looking at the driver's code seems to be an issue that affects all servers.

ennuite:
## What's Changed

Instead of always obtaining a result set for queries issued via
PreparedStatement.execute(), we now inspect the dataset_schema
returned in ActionCreatePreparedStatementResult. If the schema has no
fields, we retrieve the update count instead. This aligns the return
value with the expectations of the JDBC API.

For such cases, the Arrow Flight SQL path now uses
CommandPreparedStatementUpdate instead of
CommandPreparedStatementQuery. This change mirrors the existing
approach in Statement.execute() and Statement.executeUpdate().

### Are these changes tested?

Yes

Closes #797.

### DIFF
--- a/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightMetaImpl.java
+++ b/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightMetaImpl.java
@@ -62,14 +62,17 @@ public class ArrowFlightMetaImpl extends MetaImpl {
         parameterSchema == null
             ? new ArrayList<>()
             : ConvertUtils.convertArrowFieldsToAvaticaParameters(parameterSchema.getFields());
-
+    StatementType statementType =
+        resultSetSchema == null || resultSetSchema.getFields().isEmpty()
+            ? StatementType.IS_DML
+            : StatementType.SELECT;
     return new Signature(
         columnMetaData,
         sql,
         parameters,
         Collections.emptyMap(),
         null, // unnecessary, as SQL requests use ArrowFlightJdbcCursor
-        StatementType.SELECT);
+        statementType);
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


lidavidm (commented):
Thanks, this looks reasonable to me.

alamb:
> @alamb @jbonofre do InfluxDB and Dremio satisfy the assumptions here (dataset_schema will be populated if and only if the query of a prepared statement returns a result set)? I'm wondering if there might be systems that don't analyze the query at first and always omit the schema until actual execution

Yes.

However, I think all our queries return results (InfluxDB doesn't do DDL / DML using SQL -- there is an alternate write path) so this particular issue is unlikely to affect us

Our system will actually analyze / plan the query twice -- once on the prepare and once when executing. While this is somewhat wasteful it allows us to remain stateless between requests

PR #824, opened 2025-08-22
Make UUID implementation available for downstream projects

### LINKED ISSUE #823: Make UUID extension type publicly available

This issue proposes to make the UUID extension type implementation publicly available to users of the Arrow Java library. Currently, the UUID type, vector, and holder implementations as well as the writer and reader implementations exist in the test directory, possibly intended as an example on `ExtensionType` usage.

This would eliminating the need for users to implement their own UUID extension type around the lack of built-in UUID data type support.

bodduv:
## What's Changed

This PR moves the UUID implementation from `tests` to appropriate locations so downstream projects can use them. This allows to avoid downstream projects to have their own custom implementation for a well-understood data type. A few modifications, such as the use `ArrowBuf`, are proposed.

Closes #823 .


lidavidm (changes requested):
The UUID type here was never meant to be a real type, it was implemented solely for testing things and doesn't match the UUID canonical extension type. If there's a desire for a UUID type, please open a new issue and implement the canonical extension type instead.

lidavidm:
It's possible they are the same, but at the very least the extension name here is wrong, and I would prefer it be clearly documented and in its own namespace, and for everything to be reviewed with that in mind instead of just moving test code into a feature that has to be supported long-term.

PR #840, opened 2025-09-07
GH-839: Fix support for ResultSet.getObject for TIMESTAMP_WITH_TIMEZONE

### LINKED ISSUE #839: ResultSet.getObject throws "not implemented" exception for TIMESTAMP_WITH_TIMEZONE

TIMESTAMP_WITH_TIMEZONE support was added in https://github.com/apache/arrow-java/pull/464. However, I recently found that calling `ResultSet.getObject(int columnIndex)` throws the following exception:

```
java.lang.RuntimeException: not implemented

	at org.apache.calcite.avatica.AvaticaSite.notImplemented(AvaticaSite.java:422)
	at org.apache.calcite.avatica.AvaticaSite.get(AvaticaSite.java:381)
	at org.apache.calcite.avatica.AvaticaResultSet.getObject(AvaticaResultSet.java:395)
```

The issue is that `AvaticaSite` does not implement support for it. I will put up a PR to add support by overriding our `ResultSet.getObject` method.

aiguofer:
## What's Changed

Turns out AvaticaSite.get does not account for TIMESTAMP_WITH_TIMEZONE types so we add support on an override function.

Closes #839.

### DIFF
--- a/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightJdbcVectorSchemaRootResultSet.java
+++ b/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightJdbcVectorSchemaRootResultSet.java
@@ -19,6 +19,7 @@ package org.apache.arrow.driver.jdbc;
 import java.sql.ResultSet;
 import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
+import java.sql.Types;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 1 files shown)


aiguofer:
This PR is missing tests. I'm not sure the best place to add them, but if someone has a suggestion that'd be great!

flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightJdbcVectorSchemaRootResultSet.java
```diff
@@ -102,6 +106,33 @@ void populateData(final VectorSchemaRoot vectorSchemaRoot, final Schema schema)
...
+   * @throws SQLException
+   */
+  @Override
+  public Object getObject(int columnIndex) throws SQLException {
+    this.checkOpen();
+
+    Cursor.Accessor accessor;
+    try {
+      accessor = accessorList.get(columnIndex - 1);
+    } catch (IndexOutOfBoundsException var3) {
```
lidavidm, line 124: Nit but why `var3` and not `e`, `exc`, etc.? (This wasn't copied from a decompiler was it?)

lidavidm (commented):
Just one last question, thanks

flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightJdbcVectorSchemaRootResultSet.java
aiguofer: @lidavidm btw, this has been fixed already

lidavidm: Thanks for the ping, sorry, I was out at back to back conferences and I'm still getting back into things :sweat_smile:

lidavidm (approved):

PR #868, opened 2025-10-03
macos-13 is deprecated

### LINKED ISSUE #869: replace macos-13 github runner

macos-13 is deprecated
https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/

kevinjqliu:
## What's Changed
Closes #869

https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/

- update use of `macos-13`, replacing with `macos-15-intel` based on
https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories
- revert [previous commit](https://github.com/kevinjqliu/arrow-java/commit/abef7af8490d706d52821e56afc6d1eb21e24faf) specifically for fixing `macos-13`

### DIFF
--- a/.github/workflows/rc.yml
+++ b/.github/workflows/rc.yml
@@ -221,6 +221,10 @@ jobs:
           # llvm@14 because llvm is newer than llvm@14.
           brew uninstall llvm || :
 
+          # We can remove this when we drop support for
+          # macos-15-intel. because macos-14 or later with arm64 uses /opt/homebrew/
+          # not /usr/local/.
+          #
           # Ensure updating python@XXX with the "--overwrite" option.
           # If python@XXX is updated without "--overwrite", it causes
           # a conflict error. Because Python 3 installed not by
@@ -229,9 +233,10 @@ jobs:
           # tries to replace /usr/local/bin/2to3 and so on and causes
           # a conflict error.
           brew update
-          for python_package in $(brew list | grep python@); do
+          for python_package in $(brew list | grep python@ | sort -r); do
             brew install --overwrite ${python_package}
           done
+          brew install --overwrite python3
 
           if [ "$(uname -m)" = "arm64" ]; then
             # pkg-config formula is deprecated but it's still installed


.github/workflows/rc.yml
```diff
@@ -168,7 +168,7 @@ jobs:
       fail-fast: false
       matrix:
         platform:
-          - { runs_on: macos-13, arch: "x86_64"}
+          - { runs_on: macos-latest, arch: "aarch_64"}
```
kou, line 171: We can remove this pattern entirely.

We just drop support for Intel macOS.

.github/workflows/rc.yml
```diff
@@ -233,10 +229,10 @@ jobs:
           # tries to replace /usr/local/bin/2to3 and so on and causes
           # a conflict error.
           brew update
-          for python_package in $(brew list | grep python@ | sort -r); do
+          for python_package in $(brew list | grep python@); do
             brew install --overwrite ${python_package}
           done
-          brew install --overwrite python3
+          brew install --overwrite python
```
kou, line 235: We can remove this.

.github/workflows/rc.yml
```diff
@@ -233,10 +229,10 @@ jobs:
...
 
           if [ "$(uname -m)" = "arm64" ]; then
```
kou, line 237: We can remove this `if`.

.github/workflows/rc.yml
kou, line 242: We may be able to remove this entirely now.

kou (commented):
Could you open an issue for this?

Could you also remove `macos-13` job in `test.yml`?

Could you also update the `macosx-x86_64` part in https://github.com/apache/arrow-java/blob/main/docs/source/cdata.rst ?

.github/workflows/rc.yml
kevinjqliu: i made the change to replace macos-13 with macos-15-intel per https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories

kevinjqliu:
thanks for the review @kou
im trying to keep this pr scoped to the minimum necessary to replace macos-13. seems like macos-15-intel is the direct replacement according to https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories

.github/workflows/rc.yml
```diff
@@ -233,10 +229,9 @@ jobs:
           # tries to replace /usr/local/bin/2to3 and so on and causes
           # a conflict error.
           brew update
-          for python_package in $(brew list | grep python@ | sort -r); do
+          for python_package in $(brew list | grep python@); do
             brew install --overwrite ${python_package}
           done
-          brew install --overwrite python3
```
kou, line 239: If we keep Intel macOS support, we should not remove these changes because `macos-15-intel` still uses `/usr/local` as Homebrew prefix: https://github.com/apache/arrow-java/actions/runs/18250044030/job/51963266710?pr=868#step:9:65

```text
To restore the stashed changes to /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core, run:
```

PR #892, opened 2025-10-22
GH-891: Add ExtensionTypeWriterFactory to TransferPair

### LINKED ISSUE #891: Add ExtensionTypeWriterFactory to TransferPair

When copying data between `ListVector` instances that contain extension types using `copyFromSafe`, the extension type metadata is not properly preserved. The current implementation doesn't pass the `ExtensionTypeWriterFactory` down to the `TransferPair`, which is needed to correctly serialize extension types.

jhrotko:
## What's Changed

This PR simplifies extension type writer creation by moving from a factory-based pattern to a type-based pattern. Instead of passing `ExtensionTypeWriterFactory` instances through multiple API layers, extension types now provide their own writers via a new `getNewFieldWriter()` method on `ArrowType.ExtensionType`.

- Added `getNewFieldWriter(ValueVector)` abstract method to `ArrowType.ExtensionType`
- Removed `ExtensionTypeWriterFactory` interface and all implementations
- Removed factory parameters from `ComplexCopier`, `PromotableWriter`, and `TransferPair` APIs
- Updated `UnionWriter` to support extension types (previously threw `UnsupportedOperationException`)
- Simplified extension type implementations (`UuidType`, `OpaqueType`)

The factory pattern didn't scale well. Each new extension type required creating a separate factory class and passing it through multiple API layers. This was especially painful for external developers who had to maintain two classes per extension type and manage factory parameters everywhere.

The new approach follows the same pattern as `MinorType`, where each type knows how to create its own writer. This reduces boilerplate, simplifies the API, and makes it easier to implement custom extension types outside arrow-java.

## Breaking Changes

- `ExtensionTypeWriterFactory` has been removed
- Extension types must now implement `getNewFieldWriter(ValueVector vector)` method
- ExtensionHolders must implement `type()` which returns the `ExtensionType` for that Holder
- (Writers are obtained directly from the extension type, not from a factory)

### Migration Guide

- _Extension types must now implement `getNewFieldWriter(ValueVector vector)` method_
```java
public class UuidType extends ExtensionType {
  ...

  @Override
  public FieldWriter getNewFieldWriter(ValueVector vector) {
    return new UuidWriterImpl((UuidVector) vector);
  }
  ...
}
```
- _ExtensionHolders must implement `type()` which returns the `ExtensionType` for that Holder_
```java
public class UuidHolder extends ExtensionHolder {
   ...

  @Override
  public ArrowType type() {
    return UuidType.INSTANCE;
  }
```

- How to use Extension Writers?
  **Before:**
  ```java
    writer.extension(UuidType.INSTANCE);
    writer.addExtensionTypeWriterFactory(extensionTypeWriterFactory);
    writer.writeExtension(value);
  ```

  **After:**
  ```java
    writer.extension(UuidType.INSTANCE);
    writer.writeExtension(value);
  ```
- Also `copyAsValue` does not need to provide the factory anymore.

Closes #891 .

### DIFF
--- a/vector/src/main/java/org/apache/arrow/vector/complex/impl/AbstractExtensionTypeWriter.java
+++ b/vector/src/main/java/org/apache/arrow/vector/complex/impl/AbstractExtensionTypeWriter.java
@@ -17,8 +17,6 @@
 package org.apache.arrow.vector.complex.impl;
 
 import org.apache.arrow.vector.ExtensionTypeVector;
-import org.apache.arrow.vector.ValueVector;
-import org.apache.arrow.vector.complex.writer.FieldWriter;
 import org.apache.arrow.vector.types.pojo.Field;
 
 /**
--- a/vector/src/main/java/org/apache/arrow/vector/complex/impl/UuidReaderImpl.java
+++ b/vector/src/main/java/org/apache/arrow/vector/complex/impl/UuidReaderImpl.java
@@ -61,6 +61,7 @@ public class UuidReaderImpl extends AbstractFieldReader {
     return !vector.isNull(idx());
   }
 
+  @Override
   public void read(ExtensionHolder holder) {
     if (holder instanceof UuidHolder) {
       vector.get(idx(), (UuidHolder) holder);
--- a/vector/src/main/java/org/apache/arrow/vector/complex/impl/UuidWriterImpl.java
+++ b/vector/src/main/java/org/apache/arrow/vector/complex/impl/UuidWriterImpl.java
@@ -18,8 +18,6 @@ package org.apache.arrow.vector.complex.impl;
 
 import org.apache.arrow.memory.ArrowBuf;
 import org.apache.arrow.vector.UuidVector;
-import org.apache.arrow.vector.ValueVector;
-import org.apache.arrow.vector.complex.writer.FieldWriter;
 import org.apache.arrow.vector.holders.ExtensionHolder;
 import org.apache.arrow.vector.holders.NullableUuidHolder;
 import org.apache.arrow.vector.holders.UuidHolder;


jhrotko:
Hello, @lidavidm! Could you take a look at this PR? Also, I don't have permissions to change the label

vector/src/main/java/org/apache/arrow/vector/util/TransferPairWithExtendedType.java
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.arrow.vector.util;
+
+import org.apache.arrow.vector.complex.impl.ExtensionTypeWriterFactory;
+
+public interface TransferPairWithExtendedType extends TransferPair {
```
laurentgo: (nit) public interfaces should have proper javadoc

vector/src/main/java/org/apache/arrow/vector/complex/LargeListVector.java
```diff
@@ -661,6 +667,11 @@ public TransferImpl(Field field, BufferAllocator allocator, CallBack callBack) {
       this(new LargeListVector(field, allocator, callBack));
     }
 
+    public TransferImpl(LargeListVector to, ExtensionTypeWriterFactory writerFactory) {
```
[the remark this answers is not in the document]
jhrotko: This solution would break some cases in Dremio. With this commit design we would need to check if it is necessary a factory or not before calling the copyfrom, because we would hit a UnsupportedException for BaseValueVectors. Ideally this extension factory shouldn't exist at this level. We have Timestamp, Uuid and Variant in Dremio that are new Extension types and the implementations should "live" in arrow instead and we need to pass the DremioExtensionFactory while doing this transfer with the current design. I am also wondering if this factory responsability should be from the extension type itself and not for all types. I am still thinking about a better solution but I am also a bit reluctant to change the current design/API

vector/src/main/codegen/templates/ComplexCopier.java
```diff
@@ -41,15 +41,8 @@ public class ComplexCopier {
    * @param input field to read from
    * @param output field to write to
    */
-  public static void copy(FieldReader input, FieldWriter output) {
-    writeValue(input, output, null);
-  }
-
-  public static void copy(FieldReader input, FieldWriter output, ExtensionTypeWriterFactory extensionTypeWriterFactory) {
```
laurentgo: Removing this method would be a backward incompatible change, it may be better to actually deprecate it and documenting the pattern, to be removed at a later time (see `ComplexWriters` for example)

vector/src/main/codegen/templates/AbstractFieldReader.java
```diff
@@ -109,10 +109,6 @@ public void copyAsField(String name, ${name}Writer writer) {
 
   </#list></#list>
 
-  public void copyAsValue(StructWriter writer, ExtensionTypeWriterFactory writerFactory) {
```
laurentgo: I guess this is okay to remove a public method because there has been no release yet?

Maybe we should discuss it on the mailing list as it seems we haven't found the right pattern yet

jhrotko:
# Design Analysis: Extension Type Writer Implementation

## Executive Summary

This PR represents a significant architectural improvement in how Arrow Java handles extension type writers. The change moves from a **factory-based pattern** to a **type-based pattern**, where the `ArrowType.ExtensionType` itself provides the writer implementation. This simplifies the API, improves extensibility, and makes it easier to implement custom extension types outside of arrow-java.

---

## Problem Statement

### The Original Challenge

In Arrow's type system, each `MinorType` (INT, FLOAT, VARCHAR, etc.) has a corresponding writer implementation. However, **extension types** are special:

- They all share the same `MinorType.EXTENSIONTYPE`
- But each extension type (UUID, Opaque, custom types) needs a **different writer implementation**
- The system needed a way to map from a specific extension type to its writer

### Previous Solution: Factory Pattern

The original implementation (commits `34060eb4`, `7a7e4edd`, `7fe36d70`, `8663ffc6`) used an `ExtensionTypeWriterFactory` pattern:

```java
// Old approach - Factory pattern
public interface ExtensionTypeWriterFactory {
    FieldWriter createWriter(ValueVector vector);
}

// Usage in ComplexCopier
writer.addExtensionTypeWriterFactory(extensionTypeWriterFactory);
writer.writeExtension(value);
```

**How it worked:**
1. Each extension type had a separate factory class (e.g., `UuidWriterFactory`)
2. The factory was passed around as a parameter
3. Writers stored factories and used them to create the appropriate writer
4. TransferPair implementations also needed to carry factories

---

## Problems with the Factory Pattern

### 1. **Scalability Issues**

- Every new extension type required creating a new factory class
- Factories had to be passed through multiple layers of the API
- TransferPair implementations became complex with factory parameters

### 2. **External Extension Types**

For developers implementing extension types **outside of arrow-java**:

```java
// Developer needs to create TWO classes:
class MyCustomType extends ExtensionType { ... }
class MyCustomWriterFactory implements ExtensionTypeWriterFactory { ... }

// And manage both separately
```

This created unnecessary complexity and coupling.

### 3. **API Complexity**

Methods proliferated with factory parameters:

```java
// Old API
ComplexCopier.copy(reader, writer, extensionTypeWriterFactory);
writer.addExtensionTypeWriterFactory(factory);
TransferPair.makeTransferPair(target, factory);
```

### 4. **Tight Coupling**

The factory pattern created tight coupling between:
- The type definition
- The writer implementation  
- The factory that connects them
- All the code that needs to pass factories around

---

## New Solution: Type-Based Pattern

### Core Design Change

Add a single abstract method to `ArrowType.ExtensionType`:

```java
public abstract class ExtensionType extends ArrowType {
    // NEW METHOD
    public abstract FieldWriter getNewFieldWriter(ValueVector vector);
    
    // Existing methods
    public abstract ArrowType storageType();
    public abstract String extensionName();
    public abstract FieldVector getNewVector(...);
    // ...
}
```

### Implementation Example

```java
public class UuidType extends ExtensionType {
    @Override
    public FieldWriter getNewFieldWriter(ValueVector vector) {
        return new UuidWriterImpl((UuidVector) vector);
    }
    
    // Other methods...
}
```

---

## Benefits of the New Design

### 1. **Simplicity**

- **One class per extension type** instead of two (type + factory)
- The type itself knows how to create its writer
- Natural object-oriented design

### 2. **Cleaner API**

```java
// New API - no factory parameters needed
ComplexCopier.copy(reader, writer);
writer.writeExtension(value, type);

// The type provides the writer internally
FieldWriter writer = extensionType.getNewFieldWriter(vector);
```

### 3. **Better Extensibility**

External developers only need to implement one class:

```java
public class MyCustomType extends ExtensionType {
    @Override
    public FieldWriter getNewFieldWriter(ValueVector vector) {
        return new MyCustomWriterImpl((MyCustomVector) vector);
    }
    
    // Implement other required methods...
}
```

### 4. **Consistency with Existing Patterns**

This mirrors how `MinorType` already works:

```java
// MinorType enum (existing pattern)
public enum MinorType {
    INT(new Int(...)) {
        @Override
        public FieldWriter getNewFieldWriter(ValueVector vector) {
            return new IntWriterImpl((IntVector) vector);
        }
    },
    // ...
}

// ExtensionType (new pattern - same approach)
public class UuidType extends ExtensionType {
    @Override
    public FieldWriter getNewFieldWriter(ValueVector vector) {
        return new UuidWriterImpl((UuidVector) vector);
    }
}
```

### 5. **Decoupling**

- Writers no longer need to store or manage factories
- TransferPair implementations simplified
- Type information flows naturally through the `ArrowType` object

---

## Key Changes in the Codebase

### 1. **ArrowType.ExtensionType** (Core Change)

```java
// Added abstract method
public abstract FieldWriter getNewFieldWriter(ValueVector vector);
```

### 2. **Extension Type Implementations**

**UuidType:**
```java
@Override
public FieldWriter getNewFieldWriter(ValueVector vector) {
    return new UuidWriterImpl((UuidVector) vector);
}
```

**OpaqueType:**
```java
@Override
public FieldWriter getNewFieldWriter(ValueVector vector) {
    throw new UnsupportedOperationException("WriterImpl not yet implemented.");
}
```

### 3. **UnionWriter** (Now Supports Extension Types)

```java
// OLD: Threw UnsupportedOperationException
private ExtensionWriter getExtensionWriter(ArrowType arrowType) {
    throw new UnsupportedOperationException("ExtensionTypes are not supported yet.");
}

// NEW: Uses type to get writer
private Map<ArrowType, ExtensionWriter> extensionWriters = new HashMap<>();

private ExtensionWriter getExtensionWriter(ArrowType arrowType) {
    ExtensionWriter w = extensionWriters.get(arrowType);
    if (w == null) {
        w = ((ExtensionType) arrowType).getNewFieldWriter(data.getExtension(arrowType));
        w.setPosition(idx());
        extensionWriters.put(arrowType, w);
    }
    return w;
}
```

### 4. **ComplexCopier** (Simplified)

```java
// OLD: Required factory parameter
case EXTENSIONTYPE:
    if (extensionTypeWriterFactory == null) {
        throw new IllegalArgumentException("Must provide ExtensionTypeWriterFactory");
    }
    writer.addExtensionTypeWriterFactory(extensionTypeWriterFactory);
    writer.writeExtension(value);
    break;

// NEW: Type provides writer
case EXTENSIONTYPE:
    if (reader.isSet()) {
        Object value = reader.readObject();
        if (value != null) {
            writer.writeExtension(value);
        }
    }
    break;
```

### 5. **PromotableWriter** (Simplified)

```java
// OLD: Used UnionExtensionWriter
case EXTENSIONTYPE:
    writer = new UnionExtensionWriter((ExtensionTypeVector) vector);
    break;

// NEW: Type provides writer
case EXTENSIONTYPE:
    writer = ((ExtensionType) vector.getField().getType()).getNewFieldWriter(vector);
    break;
```

---

## Migration Impact

### For Arrow Java Developers

- **Removed:** `ExtensionTypeWriterFactory` interface and implementations
- **Removed:** Factory parameters from APIs
- **Added:** `getNewFieldWriter()` method to extension types

### For External Extension Type Developers

**Before:**
```java
// Two classes needed
class MyType extends ExtensionType { ... }
class MyWriterFactory implements ExtensionTypeWriterFactory { ... }
```

**After:**
```java
// One class needed
class MyType extends ExtensionType {
    @Override
    public FieldWriter getNewFieldWriter(ValueVector vector) {
        return new MyWriterImpl((MyVector) vector);
    }
}
```

jhrotko:
@lidavidm @xxlaykxx could you also take a look?

lidavidm (commented):
From a brief glance this approach seems more reasonable

vector/src/main/codegen/templates/PromotableWriter.java
```diff
@@ -540,18 +543,25 @@ public void writeLargeVarChar(String value) {
     getWriter(MinorType.LARGEVARCHAR).writeLargeVarChar(value);
   }
 
+  protected ArrowType lastExtensionType;
+
   @Override
   public void writeExtension(Object value) {
```
laurentgo: (design) should we deprecate this method? (since we now have `writeExtension(ExtensionHolder)`

lidavidm:
@jarohen does XTDB use extension types?

vector/src/main/codegen/templates/AbstractFieldReader.java
jhrotko: I will resolve this comment in favor to the most recent [one](https://github.com/apache/arrow-java/pull/892#discussion_r2512505808)

lidavidm:
sorry, missed this...kicked off the CI

jhrotko:
@lidavidm I see that the CI failures are unrelated to the changes and other PRs are having the same issues

jbonofre (approved):
LGTM

jhrotko:
@jbonofre @lidavidm @laurentgo @xxlaykxx thank you so much for the reviews and support!

PR #903, opened 2025-11-03
GH-825: Add UUID canonical extension type

### LINKED ISSUE #825: Introduce type support for UUID as a canonical extension type

Following standardization rules for the canonical extension type, this issue proposes to introduce UUID type support. UUID as a canonical extension type is mentioned in https://arrow.apache.org/docs/format/CanonicalExtensions.html#uuid.

jhrotko:
## What's Changed

Add Uuid Extension type canonical implementation in Java. The proposed implementation was validated in Dremio engine

Closes #825

### DIFF
--- a/vector/src/main/java/org/apache/arrow/vector/extension/UuidType.java
+++ b/vector/src/main/java/org/apache/arrow/vector/extension/UuidType.java
@@ -16,7 +16,6 @@
  */
 package org.apache.arrow.vector.extension;
 
-import java.util.concurrent.atomic.AtomicBoolean;
 import org.apache.arrow.memory.BufferAllocator;
 import org.apache.arrow.vector.FieldVector;
 import org.apache.arrow.vector.FixedSizeBinaryVector;
@@ -64,19 +63,8 @@ public class UuidType extends ExtensionType {
   /** Storage type for UUID: FixedSizeBinary(16). */
   public static final ArrowType STORAGE_TYPE = new ArrowType.FixedSizeBinary(UUID_BYTE_WIDTH);
 
-  /** Registers the UuidType in the extension type registry. */
-  private static final AtomicBoolean registered = new AtomicBoolean(false);
-
   static {
     ExtensionTypeRegistry.register(INSTANCE);
-  }
-
-  /** Register the extension type so it can be used globally. */
-  public static void ensureRegistered() {
-    if (!registered.getAndSet(true)) {
-      // The values don't matter, we just need an instance
-      ExtensionTypeRegistry.register(UuidType.INSTANCE);
-    }
   }
 
   @Override
--- a/vector/src/test/java/org/apache/arrow/vector/TestUtils.java
+++ b/vector/src/test/java/org/apache/arrow/vector/TestUtils.java
@@ -20,6 +20,7 @@ import java.util.Random;
 import org.apache.arrow.memory.BufferAllocator;
 import org.apache.arrow.vector.types.Types.MinorType;
 import org.apache.arrow.vector.types.pojo.ArrowType;
+import org.apache.arrow.vector.types.pojo.ExtensionTypeRegistry;
 import org.apache.arrow.vector.types.pojo.FieldType;
 
 public class TestUtils {
@@ -61,5 +62,15 @@ public class TestUtils {
       sb.append(random.nextInt(10)); // 0-9
     }
     return sb.toString();
+  }
+
+  /*
+   * Ensure the extension type is registered, as there might other tests trying to unregister the
+   * type. ex.: TestExtensionType#readUnderlyingType
+   */
+  public static void ensureRegistered(ArrowType.ExtensionType type) {
+    if (ExtensionTypeRegistry.lookup(type.extensionName()) == null) {
+      ExtensionTypeRegistry.register(type);
+    }
   }
 }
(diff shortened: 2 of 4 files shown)


jhrotko:
@lidavidm could you take another look? Let me know if anything else is missing

vector/src/main/java/org/apache/arrow/vector/extension/UuidType.java
```diff
@@ -14,24 +14,37 @@
...
 import org.apache.arrow.memory.BufferAllocator;
 import org.apache.arrow.vector.FieldVector;
 import org.apache.arrow.vector.FixedSizeBinaryVector;
 import org.apache.arrow.vector.UuidVector;
+import org.apache.arrow.vector.types.pojo.ArrowType;
 import org.apache.arrow.vector.types.pojo.ArrowType.ExtensionType;
+import org.apache.arrow.vector.types.pojo.ExtensionTypeRegistry;
+import org.apache.arrow.vector.types.pojo.FieldType;
 
 public class UuidType extends ExtensionType {
```
bodduv: A short doc string mentioning UuidType as a canonical extension type could make it more readable.

vector/src/main/java/org/apache/arrow/vector/extension/UuidType.java
```diff
@@ -14,24 +14,37 @@
...
+  public static final UuidType INSTANCE = new UuidType();
+
+  public static final String EXTENSION_NAME = "arrow.uuid";
+  public static final int UUID_BYTE_WIDTH = 16;
+  public static final int UUID_STRING_WIDTH = 36;
+  public static final ArrowType STORAGE_TYPE = new ArrowType.FixedSizeBinary(UUID_BYTE_WIDTH);
+
+  static {
+    ExtensionTypeRegistry.register(INSTANCE);
+  }
```
bodduv: Should the same pattern be used as in `OpaqueType.ensureRegistered`?

vector/src/main/java/org/apache/arrow/vector/FixedSizeExtensionType.java
lidavidm: Is this really necessary?

vector/src/main/java/org/apache/arrow/vector/UuidVector.java
```diff
@@ -0,0 +1,320 @@
...
+      holder.isSet = 0;
+    } else {
+      holder.isSet = 1;
+      holder.buffer = getBufferSlicePostNullCheck(index);
+    }
+  }
+
+  public void get(int index, UuidHolder holder) {
+    if (NullCheckingForGet.NULL_CHECKING_ENABLED && this.isSet(index) == 0) {
+      holder.isSet = 0;
```
lidavidm: Isn't the point of this holder that it can't be null?

vector/src/main/java/org/apache/arrow/vector/extension/UuidType.java
jhrotko: I am not sure, I see the method is only being used in tests for opaquetype. Let me add tests for UuidType as well

vector/src/main/java/org/apache/arrow/vector/UuidVector.java
jhrotko: nice catch!

vector/src/main/java/org/apache/arrow/vector/FixedSizeExtensionType.java
lidavidm: I think the problem then is that the method shouldn't be on an abstract class, it should be on an interface that isn't extension-type specific, right?

jhrotko: so I moved the method `getTypeWidth` to `FixedWidthVector` and removed `FixedSizeExtensionType` in [b04af47](https://github.com/apache/arrow-java/pull/903/commits/b04af4748b70da86e2e3d1451c659b2b92581069). This is ok because `BaseWidthVector` implements `FixedWidthVector` and `UuidVector` as well.

lidavidm:
It's https://github.com/apache/arrow-go/pull/571

jhrotko:
@lidavidm it was merged, should this be re-run?

lidavidm (commented):
This has conflicts

vector/src/main/java/org/apache/arrow/vector/extension/UuidType.java
```diff
@@ -0,0 +1,121 @@
...
+
+  /** Register the extension type so it can be used globally. */
+  public static void ensureRegistered() {
+    if (!registered.getAndSet(true)) {
+      // The values don't matter, we just need an instance
+      ExtensionTypeRegistry.register(UuidType.INSTANCE);
+    }
+  }
```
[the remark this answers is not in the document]
jhrotko: this is only used in tests similarly to Opaque extension type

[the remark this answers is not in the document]
lidavidm: There is no static initializer there. There is one here.

lidavidm (approved):

jbonofre (approved):

lidavidm:
I think these failures are related to upstream and won't go away on rebase.

jbonofre:
I will check as I try to stabilize our CI. It fails way too often and it's an issue in PRs.

PR #910, opened 2025-11-18
GH-863: [JDBC] Suppress benign exceptions from gRPC layer on ArrowFlightSqlClientHandler#close

### LINKED ISSUE #863: ArrowFlightSqlClientHandler.close() performs a CloseSession RPC that can fail during gRPC channel shutdown - similar to ARROW-17785

Just like in this issue:
https://github.com/apache/arrow/issues/33012

there are still codepaths that trigger a faulty shutdown when using CATALOG_ID with tools like JMeter which outputs a message like this one and prevents the test run from succeeding. 

> Thread Name:Warmup 1-1
> Sample Start:2025-09-12 15:28:27 IST
> Load time:1060
> Connect Time:1060
> Latency:0
> Size in bytes:66
> Sent bytes:0
> Headers size in bytes:0
> Body size in bytes:66
> Sample Count:1
> Error Count:1
> Data type ("text"|"bin"|""):text
> Response code:null 0
> Response message:java.sql.SQLException: Cannot create PoolableConnectionFactory (Channel shutdown invoked)
> 
> 
> SampleResult fields:
> ContentType: text/plain
> DataEncoding: UTF-8

I'll submit a PR with a similar approach to:
https://github.com/apache/arrow/pull/14210/commits

for the additional code path.

ennuite:
## What's Changed

When using the Flight SQL JDBC driver with connection pooling and a catalog parameter, ArrowFlightSqlClientHandler.close() performs a CloseSession RPC that can fail during gRPC channel shutdown. 

These transient failures (UNAVAILABLE or INTERNAL with "Connection closed after GOAWAY") cause noisy errors in pooling frameworks like Apache Commons DBCP. 

With this PR these exceptions will instead be suppressed and logged, following the procedure that was used for [ARROW-17785](https://issues.apache.org/jira/browse/ARROW-17785) 

### Are these changes tested?

Yes

Closes #863

### DIFF
--- a/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/client/ArrowFlightSqlClientHandler.java
+++ b/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/client/ArrowFlightSqlClientHandler.java
@@ -338,7 +338,7 @@ public final class ArrowFlightSqlClientHandler implements AutoCloseable {
   private void logSuppressedCloseException(
       FlightRuntimeException fre, String operationDescription) {
     // ARROW-17785 and GH-863: suppress exceptions caused by flaky gRPC layer during shutdown
-    LOGGER.info("Suppressed error {}", operationDescription, fre);
+    LOGGER.debug("Suppressed error {}", operationDescription, fre);
   }
 
   /** A prepared statement handler. */


flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/client/ArrowFlightSqlClientHandler.java
```diff
@@ -262,15 +262,85 @@ public FlightInfo getInfo(final String query) {
...
+  /**
+   * Logs a suppressed close exception with appropriate level based on debug settings.
+   *
+   * @param fre the FlightRuntimeException being suppressed
+   * @param operationDescription description of the operation for logging
+   */
+  private void logSuppressedCloseException(
+      FlightRuntimeException fre, String operationDescription) {
+    // ARROW-17785 and GH-863: suppress exceptions caused by flaky gRPC layer during shutdown
+    LOGGER.info("Suppressed error {}", operationDescription, fre);
```
lidavidm: I would still prefer debug level

jbonofre: Agreed. Debug makes more sense to me here.

lidavidm (approved):

PR #938, opened 2025-12-24
GH-399: Check for null writers in DenseUnionWriter#setPosition

### LINKED ISSUE #399: [Java] DenseUnionWriter#setPosition fails with NullPointerException
The writer always iterates through all BaseWriters, and an array of 128 BaseWriters is allocated. So if you do not have 128 typeIds and do not touch all of them, setPosition will give you an exception.

**Reporter**: [David Li](https://issues.apache.org/jira/browse/ARROW-8665) / @lidavidm

<sub>**Note**: *This issue was originally created as [ARROW-8665](https://issues.apache.org/jira/browse/ARROW-8665). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

Kaustav-Sarkar:
## GH-399 Fix setPosition fails with NullPointerException

Fixed a `NullPointerException` in `DenseUnionWriter#setPosition`.

The issue was that `setPosition` tried to update all writers in its internal array, even if they hadn't been initialized yet. I added a null check so it only updates writers that actually exist.

Also added a regression test (`TestDenseUnionWriterNPE`) to verify the fix and updated `.gitignore`.

Closes #399.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,12 +7,12 @@
 .buildpath
 .checkstyle
 .classpath
+.cursor/
 .factorypath
 .idea/
-.vscode/
-.cursor/
 .project
 .settings/
+.vscode/
 /*-build/
 /.mvn/.develocity/
 /apache-arrow-java-*


vector/src/test/java/org/apache/arrow/vector/complex/impl/TestDenseUnionWriterNPE.java
```diff
@@ -0,0 +1,86 @@
...
+    allocator = new RootAllocator(Long.MAX_VALUE);
+  }
+
+  @AfterEach
+  public void terminate() {
+    allocator.close();
+  }
+
+  @Test
+  public void testListOfDenseUnionWriterNPE() {
```
lidavidm: Link back to the issue 
```suggestion
  public void testListOfDenseUnionWriterNPE() {
    // Regression test for ...
```

lidavidm:
Ok. The JNI test is an upstream build issue, but @Kaustav-Sarkar can you fix the format error in .gitignore? https://github.com/apache/arrow-java/actions/runs/20526378098/job/59055944634?pr=938

PR #941, opened 2025-12-29
GH-125: Allow null timestamp holder sans timezone

### LINKED ISSUE #125: [Java] IllegalArgumentException when unsetting TimeStamp*TZVector values

In apache/arrow#35352, there was a [timezone check](https://github.com/apache/arrow/blob/main/java/vector/src/main/java/org/apache/arrow/vector/TimeStampMicroTZVector.java#L164-L166) added to the `TimeStamp*TZVector.set` methods that verified that the `*TimeStamp*TZHolder` had the same timezone as the vector that was being set.

Before this change was committed, this snippet to unset the value at an index worked:
```
TimeStampMicroTZVector vector = new TimeStampMicroTZVector("vector", allocator, "UTC");
NullableTimeStampMicroTZHolder holder = new NullableTimeStampMicroTZHolder();
vector.setSafe(0, holder);
```
However, after the change, it now throws `IllegalArgumentException: holder.timezone: null not equal to vector timezone: UTC` when trying to unset the value in the vector.

We currently have a workaround for this issue by explicitly setting `holder.timezone = vector.getTimeZone();` before calling the `setSafe` method. I'm wondering if it is expected behavior to require that `holder.timezone` match the vector's timezone when unsetting the value. It also differs from the `get` behavior which [does not set the timezone on the holder when the vector's value is unset](https://github.com/apache/arrow/blob/main/java/vector/src/main/java/org/apache/arrow/vector/TimeStampMicroTZVector.java#L121-L123). I'm happy to help submit a patch if the existing behavior is indeed unintended.

### Component(s)

Java

Kaustav-Sarkar:
## Description
Fixes an `IllegalArgumentException` in `TimeStamp*TZVector.set/setSafe` when unsetting values using a holder with a `null` timezone. The validation logic now correctly ignores the timezone check when `holder.isSet <= 0`, allowing default-constructed holders to be used for unsetting values as expected.

Comprehensive tests added for all timestamp precisions (Micro, Milli, Nano, Sec) to verify the fix and ensure the existing workaround (setting explicit timezone) remains supported.

Closes #125 .


vector/src/main/java/org/apache/arrow/vector/TimeStampMicroTZVector.java
lidavidm: I think it'd be more natural to move the else block below as the second conditional check, so we aren't repeating boolean clauses.

jbonofre: I agree, it looks more "logical".

Kaustav-Sarkar:
```
Could not find a package configuration file provided by "lz4" with any of
    the following names:
  
      lz4Config.cmake
      lz4-config.cmake
  
    Add the installation prefix of "lz4" to CMAKE_PREFIX_PATH or set "lz4_DIR"
    to a directory containing one of the above files.  If "lz4" provides a
    separate development package or SDK, be sure it has been installed.
```
The build fails with this issue, I do not see how it is related to my changes. Do I need to add anything else to get this PR merged?

jbonofre:
@Kaustav-Sarkar do you mind to rebase ? It should fix the CI. Thanks !

PR #949, opened 2026-01-07
GH-948: Use buffer indexing for UUID vector

### LINKED ISSUE #948: Improve memory management in UUID vector

Improve memory management in UUID vector operations by implementing buffer indexing instead of buffer slicing. This optimization reduces memory overhead and improves performance when working with UUID values.

jhrotko:
## What's Changed

The current UUID vector implementation creates new buffer slices when reading values through holders, which has several drawbacks:
- Memory overhead: Each slice creates a new ArrowBuf object
- Performance impact: Buffer slicing is slower than direct buffer indexing
- Inconsistency: Other fixed-width types (like Decimal) use buffer indexing with a `start` offset field

### Proposed Changes

1. Add `start` field to UUID holders to track buffer offsets:
   - `UuidHolder`: Add `public int start = 0;`
   - `NullableUuidHolder`: Add `public int start = 0;`
2. Update `UuidVector` to use buffer indexing
3. Update readers and writers

### Related Work

- Original UUID extension type implementation: GH-825 (#903)

Closes #948

### DIFF
--- a/vector/src/main/java/org/apache/arrow/vector/holders/NullableUuidHolder.java
+++ b/vector/src/main/java/org/apache/arrow/vector/holders/NullableUuidHolder.java
@@ -16,9 +16,9 @@
  */
 package org.apache.arrow.vector.holders;
 
+import org.apache.arrow.memory.ArrowBuf;
 import org.apache.arrow.vector.extension.UuidType;
 import org.apache.arrow.vector.types.pojo.ArrowType;
-import org.apache.arrow.memory.ArrowBuf;
 
 /**
  * Value holder for nullable UUID values.
@@ -34,6 +34,7 @@ import org.apache.arrow.memory.ArrowBuf;
 public class NullableUuidHolder extends ExtensionHolder {
   /** Buffer containing 16-byte UUID data. */
   public ArrowBuf buffer;
+
   /** Offset in the buffer where the UUID data starts. */
   public int start = 0;
 
--- a/vector/src/test/java/org/apache/arrow/vector/TestLargeListVector.java
+++ b/vector/src/test/java/org/apache/arrow/vector/TestLargeListVector.java
@@ -37,7 +37,7 @@ import org.apache.arrow.vector.complex.impl.UnionLargeListWriter;
 import org.apache.arrow.vector.complex.reader.FieldReader;
 import org.apache.arrow.vector.complex.writer.BaseWriter.ExtensionWriter;
 import org.apache.arrow.vector.extension.UuidType;
-import org.apache.arrow.vector.holders.UuidHolder;
+import org.apache.arrow.vector.holders.NullableUuidHolder;
 import org.apache.arrow.vector.types.Types.MinorType;
 import org.apache.arrow.vector.types.pojo.ArrowType;
 import org.apache.arrow.vector.types.pojo.Field;
@@ -1071,14 +1071,14 @@ public class TestLargeListVector {
       assertTrue(reader.isSet(), "first list shouldn't be null");
       reader.next();
       FieldReader uuidReader = reader.reader();
-      UuidHolder holder = new UuidHolder();
+      NullableUuidHolder holder = new NullableUuidHolder();
       uuidReader.read(holder);
-      UUID actualUuid = UuidUtility.uuidFromArrowBuf(holder.buffer, 0);
+      UUID actualUuid = UuidUtility.uuidFromArrowBuf(holder.buffer, holder.start);
       assertEquals(u1, actualUuid);
       reader.next();
       uuidReader = reader.reader();
       uuidReader.read(holder);
-      actualUuid = UuidUtility.uuidFromArrowBuf(holder.buffer, 0);
+      actualUuid = UuidUtility.uuidFromArrowBuf(holder.buffer, holder.start);
       assertEquals(u2, actualUuid);
 
       // Verify second list
@@ -1087,12 +1087,12 @@ public class TestLargeListVector {
       reader.next();
       uuidReader = reader.reader();
       uuidReader.read(holder);
-      actualUuid = UuidUtility.uuidFromArrowBuf(holder.buffer, 0);
+      actualUuid = UuidUtility.uuidFromArrowBuf(holder.buffer, holder.start);
       assertEquals(u3, actualUuid);
       reader.next();
       uuidReader = reader.reader();
       uuidReader.read(holder);
-      actualUuid = UuidUtility.uuidFromArrowBuf(holder.buffer, 0);
+      actualUuid = UuidUtility.uuidFromArrowBuf(holder.buffer, holder.start);
       assertEquals(u4, actualUuid);
       reader.next();
       uuidReader = reader.reader();
--- a/vector/src/test/java/org/apache/arrow/vector/TestListVector.java
+++ b/vector/src/test/java/org/apache/arrow/vector/TestListVector.java
@@ -1350,14 +1350,14 @@ public class TestListVector {
       assertTrue(reader.isSet(), "first list shouldn't be null");
       reader.next();
       FieldReader uuidReader = reader.reader();
-      UuidHolder holder = new UuidHolder();
+      NullableUuidHolder holder = new NullableUuidHolder();
       uuidReader.read(holder);
-      UUID actualUuid = UuidUtility.uuidFromArrowBuf(holder.buffer, 0);
+      UUID actualUuid = UuidUtility.uuidFromArrowBuf(holder.buffer, holder.start);
       assertEquals(u1, actualUuid);
       reader.next();
       uuidReader = reader.reader();
       uuidReader.read(holder);
-      actualUuid = UuidUtility.uuidFromArrowBuf(holder.buffer, 0);
+      actualUuid = UuidUtility.uuidFromArrowBuf(holder.buffer, holder.start);
       assertEquals(u2, actualUuid);
 
       // Verify second list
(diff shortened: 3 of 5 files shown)


jbonofre:
@lidavidm yes, that's my point as well. I wonder if we should not "square" the vector (with two "edges").

lidavidm:
I don't think Varchar is an appropriate comparison here since it is variable-length. UUID is a fixed length (known at compile-time, unlike fixed-length binary) so it can be simplified. I assume both memory-wise and performance-wise, storing two longs is preferable to slicing/copying/storing a reference + length, and less prone to issues with forgetting to retain/release the buffer.

vector/src/main/java/org/apache/arrow/vector/UuidVector.java
```diff
@@ -146,17 +149,20 @@ public int isSet(int index) {
...
-  public ArrowBuf get(int index) throws IllegalStateException {
+  public void get(int index, UuidHolder holder) {
+    Preconditions.checkArgument(index >= 0, "Cannot get negative index in UUID vector.");
     if (NullCheckingForGet.NULL_CHECKING_ENABLED && this.isSet(index) == 0) {
-      throw new IllegalStateException("Value at index is null");
+      holder.isSet = 0;
     } else {
-      return getBufferSlicePostNullCheck(index);
+      holder.isSet = 1;
+      final ByteBuffer bb = ByteBuffer.wrap(getUnderlyingVector().getObject(index));
```
lidavidm: Is it not possible to directly use ArrowBuf#getLong? (Or even MemoryUtil#getLong + a manual bounds check to avoid two bounds checks, though I'd expect the JIT would optimize that out)

jhrotko:
Test Configuration
- Vector Length w/ 10,000 elements
- Old Version: Commit e35ed7c3 (UuidHolder with `ArrowBuf buffer` + `int start`)
- New Version: uuid-improvement branch (UuidHolder with `long mostSigBits` + `long leastSigBits`)
- JMH Settings: 2 warmup iterations, 3 measurement iterations, GC profiler enabled

## 1. getUuidDirectly
Read UUID objects directly via `vector.getObject(i)`

| Metric | Old (buffer+offset) | New (two-longs) | Change |
|--------|---------------------|-----------------|--------|
| **Time (µs/op)** | 10.007 ± 1.404 | 9.960 ± 0.460 |  **-0.5%** (0.047 µs faster) |
| **Memory Allocation (B/op)** | 0.012 ± 0.147 | 0.012 ± 0.145 |  **Same** |
| **Allocation Rate (MB/sec)** | 0.001 ± 0.014 | 0.001 ± 0.014 |  **Same** |
| **GC Count** | ≈ 0 | ≈ 0 | **Same** |

 --> Essentially identical performance

## 2. getWithNullableUuidHolder
Read into `NullableUuidHolder` (supports null values)

| Metric | Old (buffer+offset) | New (two-longs) | Change |
|--------|---------------------|-----------------|--------|
| **Time (µs/op)** | 9.918 ± 0.120 | 9.979 ± 0.621 | ≈ **+0.6%** (0.061 µs slower) |
| **Memory Allocation (B/op)** | 0.011 ± 0.144 | 0.011 ± 0.144 |  **Same** |
| **Allocation Rate (MB/sec)** | 0.001 ± 0.014 | 0.001 ± 0.014 |  **Same** |
| **GC Count** | ≈ 0 | ≈ 0 |  **Same** |

--> Essentially identical performance, both allocation-free

## 3. getWithUuidHolder
Read into `UuidHolder` (non-nullable)

| Metric | Old (buffer+offset) | New (two-longs) | Change |
|--------|---------------------|-----------------|--------|
| **Time (µs/op)** | 9.934 ± 0.162 | 16.488 ± 0.121 | ⚠️ **+66.0%** (6.554 µs slower) |
| **Memory Allocation (B/op)** | 0.011 ± 0.144 | 0.019 ± 0.239 |  **+73%** (still ~0 bytes) |
| **Allocation Rate (MB/sec)** | 0.001 ± 0.014 | 0.001 ± 0.014 |  **Same** |
| **GC Count** | ≈ 0 | ≈ 0 |  **Same** |

--> 66% performance regression
- Old approach: Direct buffer reference (10 µs/op)
- New approach: Read two longs + byte reversal (16 µs/op)

## 4. setUuidDirectly
Write UUID objects directly via `vector.setSafe(i, uuid)`

| Metric | Old (buffer+offset) | New (two-longs) | Change |
|--------|---------------------|-----------------|--------|
| **Time (µs/op)** | 87.047 ± 2.229 | 89.154 ± 5.116 | ≈ **+2.4%** (2.107 µs slower) |
| **Memory Allocation (B/op)** | 320000.101 ± 1.262 | 320000.103 ± 1.290 |  **Same** |
| **Allocation Rate (MB/sec)** | 3505.783 ± 88.171 | 3422.955 ± 195.204 |  **-2.4%** |
| **GC Count** | 174 | 170 |  **-2.3%** |
| **GC Time (ms)** | 79 | 78 |  **-1.3%** |

--> Essentially identical performance
- High allocation is expected (creating 10,000 UUID objects = 32 bytes each)

## 5. setWithNullableUuidHolder
Write using `NullableUuidHolder` with nullability support

| Metric | Old (buffer+offset) | New (two-longs) | Change |
|--------|---------------------|-----------------|--------|
| **Time (µs/op)** | 25.415 ± 4.776 | 25.052 ± 1.187 |  **-1.4%** (0.363 µs faster) |
| **Memory Allocation (B/op)** | 0.029 ± 0.364 | 0.029 ± 0.362 |  **Same** |
| **Allocation Rate (MB/sec)** | 0.001 ± 0.014 | 0.001 ± 0.014 |  **Same** |
| **GC Count** | ≈ 0 | ≈ 0 |  **Same** |

--> Essentially identical performance

## 6. setWithUuidHolder
Write using `UuidHolder` (non-nullable)

| Metric | Old (buffer+offset) | New (two-longs) | Change |
|--------|---------------------|-----------------|--------|
| **Time (µs/op)** |  **FAILED** | 91.831 ± 7.582 |  **FIXED** |
| **Memory Allocation (B/op)** |  **N/A** | 320000.106 ± 1.328 |  **Works** |
| **Allocation Rate (MB/sec)** |  **N/A** | 3323.156 ± 272.962 |  **Works** |
| **GC Count** |  **N/A** | 165 |  **Works** |
| **GC Time (ms)** |  **N/A** | 75 | **Works** |

**Error in Old Version**:
```
java.lang.NullPointerException: Cannot set UUID vector, the source buffer is null.
	at org.apache.arrow.vector.UuidVector.set(UuidVector.java:241)
```

--> Offset implementation was broken!
- The old buffer+offset approach had a NullPointerException bug
- The new two-longs approach fixes this bug and works correctly

## 7. setWithWriter
Write using `UuidWriterImpl` (extension writer pattern)

| Metric | Old (buffer+offset) | New (two-longs) | Change |
|--------|---------------------|-----------------|--------|
| **Time (µs/op)** | 62.174 ± 8.932 | 61.747 ± 3.117 |  **-0.7%** (0.427 µs faster) |
| **Memory Allocation (B/op)** | 213312.072 ± 0.912 | 213312.076 ± 1.036 |  **Same** |
| **Allocation Rate (MB/sec)** | 3272.006 ± 473.174 | 3294.462 ± 165.505 |  **+0.7%** |
| **GC Count** | 162 | 164 | ≈ **+1.2%** |
| **GC Time (ms)** | 73 | 76 | ≈ **+4.1%** |

--> Essentially identical performance

jhrotko:
The results of the benchmarking comparing the two versions (v1
) buffer indexing from [e35ed7c](https://github.com/apache/arrow-java/pull/949/commits/e35ed7c3585e8cf681c4e11aa9d91a7c7b041e55) and (v2) two MSB and LSB longs  UUID holder implementations across 5 scales (1k, 10k, 100k, 1M, 10M elements) for 7 different operations.

There were no major differences with exception of the test `getWithUuidHolder` where there was a 65% slow down with v2. I can provide full data for all methods if requested.

| Scale | v1 (µs/op) | v2 (µs/op) | Diff (µs) | % Change | v1/v2 ratio |
|-------|-------------|-------------|-----------|----------|---------------|
| **1k**   | 1.000 | 1.666 | +0.666 | +66.6% | 1.67x slower |
| **10k**  | 9.954 | 16.462 | +6.508 | +65.4% | 1.65x slower |
| **100k** | 102.048 | 162.946 | +60.898 | +59.7% | 1.60x slower |
| **1M**   | 1043.961 | 1669.155 | +625.194 | +59.9% | 1.60x slower |
| **10M**  | 10121.154 | 16718.845 | +6597.691 | +65.2% | 1.65x slower |

The `getWithUuidHolder` regression is due to different implementation approaches:

**V1 (ArrowBuf reference)**:
```java
holder.buffer = getDataBuffer();      // Copy pointer (8 bytes)
holder.start = getStartOffset(index); // Copy offset (4 bytes)
// Total: 12 bytes copied, NO data reading (deferred work)
```

**V2 (MSB/LSB longs)**:
```java
holder.mostSigBits = Long.reverseBytes(dataBuffer.getLong(start));      // Read + reverse
holder.leastSigBits = Long.reverseBytes(dataBuffer.getLong(start + 8)); // Read + reverse
// Total: 16 bytes READ + 2 byte reversals (immediate work)
```

The old version defers work (just stores references), while the new version does work upfront (reads and byte-reverses data -> O(N)).

---

Given the performance difference I would prefer to implement version 1 with buffer indexing in order to maintain execution performance. Other holder types already take this advantage, namely Decimal, Varchar and  Varbinary. Given that UUID takes 16 bytes it could also take advantage of the zero-copy data access design provided by these types

lidavidm:
I don't particularly mind referencing the buffer, then; I just think it's more prone to footguns, but of course it sounds like Dremio has more experience here :)

jhrotko:
@lidavidm @jbonofre can we run the CI?

jbonofre (approved):
LGTM and pretty well aligned with VarChar approach.

jhrotko:
@jbonofre @lidavidm Thank yo uso much for the reviews and the support!

PR #954, opened 2026-01-11
GH-939: Remove reflection for gRPC buffers

### LINKED ISSUE #939: [FlightRPC] Remove sketchy gRPC reflection code

https://github.com/grpc/grpc-java/pull/12579

There are public methods to get ownership of the internal ByteBuffers, instead of having to use reflection (which, did we actually ever need it in the first place?)

> There's no need to dig into our internals for accessing the ByteBuffer (which is what GetReadableBuffer reports to do). We have zero-copy APIs. They can use [HasByteBuffer.getByteBuffer()](https://grpc.github.io/grpc-java/javadoc/io/grpc/HasByteBuffer.html)+InputStream.skip() to loop through the ByteBuffers. If they want to access all the byte buffers simultaneously (as skip() will deallocate the last byte buffer, just like read() does), then they can use InputStream.mark(). If they want to take over ownership of the ByteBuffers, then they can use [Detachable.detach()](https://grpc.github.io/grpc-java/javadoc/io/grpc/Detachable.html).

xborder:
## What's Changed

- Replace reflection-based gRPC buffer handling with FlightDataParser reading from InputStream/detached ByteBuffer, enabling zero‑copy ArrowBuf slices for app_metadata and data_body.
- Simplify ArrowMessage.frame to use the shared parser and fallback path; remove GetReadableBuffer and add grpc-core runtime/module updates.
- Add TestArrowMessageParse coverage for duplicate fields, zero‑copy allocation, and non‑detach fallback behavior.

Closes #939.


lidavidm (commented):
All builds fail:

```
Error:  Non-test scoped test only dependencies found:
Error:     io.grpc:grpc-core:jar:1.78.0:compile
```

lidavidm (approved):
Approved CI

flight/flight-core/src/main/java/org/apache/arrow/flight/ArrowMessage.java
```diff
@@ -377,6 +382,114 @@ private static int readRawVarint32(int firstByte, InputStream is) throws IOExcep
...
+    // Calculate memory address accounting for buffer position
+    long baseAddress = MemoryUtil.getByteBufferAddress(detachedByteBuffer);
+    long dataAddress = baseAddress + detachedByteBuffer.position();
+
+    // Create ForeignAllocation with proper cleanup
+    ForeignAllocation foreignAllocation =
+        new ForeignAllocation(size, dataAddress) {
+          @Override
+          protected void release0() {
+            closeQuietly(detachedStream);
```
jbonofre: Don't we risk a race condition here where we close the stream (detached) before a write ?

xborder:
@jbonofre I tried to reproduce this on a Windows 11 VM and couldn't. Could this be a flaky test?

flight/flight-core/src/main/java/org/apache/arrow/flight/ArrowMessage.java
```diff
@@ -75,19 +78,10 @@ class ArrowMessage implements AutoCloseable {
     if (zeroCopyWriteFlag == null) {
       zeroCopyWriteFlag = System.getenv("ARROW_FLIGHT_ENABLE_ZERO_COPY_WRITE");
     }
-    ENABLE_ZERO_COPY_READ = !"false".equalsIgnoreCase(zeroCopyReadFlag);
+    ENABLE_ZERO_COPY_READ = true; // !"false".equalsIgnoreCase(zeroCopyReadFlag);
```
laurentgo: is it intentional?

xborder: no, fixed. thanks

PR #1002, opened 2026-02-02
MINOR: Empty commit to test CI

jbonofre:
It seems we have again something broken in the CI (JNI). This PR is to test it.

### DIFF
--- a/.env
+++ b/.env
@@ -53,4 +53,4 @@ MAVEN=3.9.9
 # Versions for various dependencies used to build artifacts
 # Keep in sync with apache/arrow
 ARROW_REPO_ROOT=./arrow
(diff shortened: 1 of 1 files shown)


jbonofre:
FYI @kou @lidavidm the CI is broker again (on JNI) due to:

```
  CMake Error at cmake_modules/ThirdpartyToolchain.cmake:313 (message):
    Couldn't find xsimd >= 14.0.0
```

I'm fixing this one.

kou:
We need to update vcpkg version. See also: https://github.com/apache/arrow/pull/49062

Or we can use Apache Arrow 23.0.0 instead of apache/arrow main.

PR #1005, opened 2026-02-05
GH-1004:  [JDBC] Fix NPE in ArrowFlightJdbcDriver#connect​(final String url, final Properties info) 

### LINKED ISSUE #1004: [JDBC] NullPointerException in DriverManager#getConnection​(String url, Properties info)

The `ArrowFlightJdbcDriver` method 
`public ArrowFlightConnection connect(final String url, final Properties info)` 
does not properly handle `null` values for `info`, resulting in an NPE.

Code for reproduction: 
 `DriverManager.getConnection(jdbcUrl, null); // where jdbcUrl is a String with a valid URL.`

ennuite:
## What's Changed
`ArrowFlightJdbcDriver.connect(final String url, final Properties info)` now properly ignores a null value for `info`, obtaining the properties solely from the URL.

Closes #1004.

### DIFF
--- a/.github/workflows/rc.yml
+++ b/.github/workflows/rc.yml
@@ -107,18 +107,12 @@ jobs:
       - name: Extract source archive
         run: |
           tar -xf apache-arrow-java-*.tar.gz --strip-components=1
-      # We always use the main branch for apache/arrow for now.
-      # Because we want to use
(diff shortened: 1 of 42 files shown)


jbonofre (approved):
That should be pretty rare, but worth to avoid this.

PR #1010, opened 2026-02-07
GH-141: Correct capacity behavior in BufferAllocator.buffer docstrings

### LINKED ISSUE #141: [Java] Allocator `buffer` method no setting the capacity correctly.

The following snippet
```java
package org.apache.arrow.vector;

import org.apache.arrow.memory.ArrowBuf;
import org.apache.arrow.memory.RootAllocator;
import org.junit.jupiter.api.Test;


public class BufferAllocatorOverAllocation {

    @Test
    void overallocation()  {
        RootAllocator allocator = new RootAllocator();
        ArrowBuf buf = allocator.buffer(9946);

        System.out.println(buf.capacity());

        buf.capacity(9946);

        System.out.println(buf.capacity());

    }

}
```
prints
```
16384
9946
```
So without setting the capacity explicitly the allocator overallocates memory and also seems to set the capacity of the buffer to that. That is contrary to the docstring of the `buffer` method:
```
Note that the buffer may technically be larger than the requested size for rounding purposes. However, the buffer's capacity will be set to the configured size.
```

Tested against the `14.0.2` tag. 

### Component(s)

Java

axreldable:
## What's Changed

Update the `BufferAllocator.buffer(long)` and `BufferAllocator.buffer(long, BufferManager)` docstrings so they match the actual behavior: the returned buffer’s capacity is the allocated (possibly rounded) size, not the requested size. The previous text said the capacity would be set to the configured size, which was incorrect. The new text also mentions that callers can use `ArrowBuf#capacity(long)` to set the capacity to the requested size when needed.

Documentation-only change; no code or behavioral changes.

Closes #141.

### DIFF
--- a/memory/memory-core/src/main/java/org/apache/arrow/memory/ArrowBuf.java
+++ b/memory/memory-core/src/main/java/org/apache/arrow/memory/ArrowBuf.java
@@ -136,7 +136,7 @@ public final class ArrowBuf implements AutoCloseable {
   /**
    * Adjusts the capacity of this buffer. Size increases are NOT supported.
    *
-   * @param newCapacity Must be in in the range [0, length).
+   * @param newCapacity Must be in the range [0, length).
    */
   public synchronized ArrowBuf capacity(long newCapacity) {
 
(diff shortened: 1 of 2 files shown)


jbonofre (approved):
That's a nit, but good one.

PR #1016, opened 2026-02-09
GH-139: [Flight] Stop return null from MetadataAdapter.getAll(String) and getAllByte(String)

### LINKED ISSUE #139: [Java][Docs] Undocumented null return from CallHeaders.getAll()

The Java CallHeaders documentation doesn't document that `.getAll()` will return `null` if no matches are found, so the reader is left with the default assumption of getting back an empty Iterator<>.

I assume this probably extends to `.getAllByte()` but I haven't had time to investigate yet.

I will probably fix the documentation in the near future, although (given it's theoretically not a breaking change—there is no guarantee of not receiving an empty Iterator<> either!) I'm somewhat more inclined to change the behaviour to eliminate the null check anywhere it's used, unless anyone objects.

### Component(s)

Java, Documentation

axreldable:
## What's Changed

`CallHeaders` has 3 implementations:
- FlightCallHeaders
- ErrorFlightMetadata
- MetadataAdapter

**Before this change:**
`MetadataAdapter` could return `null` from `getAll(String)` and `getAllByte(String)` when there were no values for the key, because gRPC’s `Metadata.getAll()` returns `null` in that case. This was undocumented and forced callers to null-check.

**After this change:**
All 3 implementations consistently return an `empty iterable` (never `null`) when the key is absent or has no values. The contract is documented on the interface and covered by tests for each implementation.

---
`MetadataAdapter.getAll(String)` and `getAllByte(String)` return empty iterator instead of null.

---
Closes #139.

### DIFF
--- a/flight/flight-core/src/main/java/org/apache/arrow/flight/grpc/MetadataAdapter.java
+++ b/flight/flight-core/src/main/java/org/apache/arrow/flight/grpc/MetadataAdapter.java
@@ -54,14 +54,16 @@ public class MetadataAdapter implements CallHeaders {
 
   @Override
   public Iterable<String> getAll(String key) {
-    final Iterable<String> all = this.metadata.getAll(Metadata.Key.of(key, Metadata.ASCII_STRING_MARSHALLER));
+    final Iterable<String> all =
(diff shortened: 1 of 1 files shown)


jbonofre:
The verify job failed to flaky test (memory allocation). I ran again the jobs to double check.

PR #1023, opened 2026-02-16
GH-1020: Update vcpkg revision

### LINKED ISSUE #1020: Update vcpkg revision for Linux JNI build

apache/arrow updated vcpkg revision: https://github.com/apache/arrow/pull/49062

kou:
## What's Changed

apache/arrow updated vcpkg revision by https://github.com/apache/arrow/pull/49062 . It's not released yet but we use apache/arrow main in our nightly CI jobs that are failed. So we should follow the change as soon as possible.

Closes #1020.

### DIFF
--- a/.env
+++ b/.env
@@ -53,4 +53,4 @@ MAVEN=3.9.9
 # Versions for various dependencies used to build artifacts
 # Keep in sync with apache/arrow
 ARROW_REPO_ROOT=./arrow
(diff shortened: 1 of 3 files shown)


kou:
Hmm. It seems that we need newer Apache Arrow C++ with newer vcpkg revision.

Let's revisit this after Apache Arrow 24.0.0 is released.

PR #1027, opened 2026-02-23
GH-577: Use JVM proxy settings in Netty client

### LINKED ISSUE #577: Flight-SQL JDBC driver ignores JVM proxy settings

Hi, I was making some tests for checking the support of proxies with the Flight-SQL JDBC driver.

GRPC supports the [JVM properties](https://docs.oracle.com/javase/8/docs/technotes/guides/net/proxies.html) for indicating the default http proxy to use. Nevertheless, I tested with Flight-SQL JDBC driver without luck. The driver is ignoring them.
I mean these ones:
`-Dhttps.proxyHost=<host> -Dhttps.proxyPort=<port>`

Checking the reason, I arrived to these lines where the Netty channel is opened:
https://github.com/apache/arrow-java/blob/bd2173c59c52983a9e1a6cadffb8345294960949/flight/flight-core/src/main/java/org/apache/arrow/flight/FlightClient.java#L854-L862

Using method `io.grpc.netty.NettyChannelBuilder#forAddress(java.net.SocketAddress)` makes GRPC to create the channel using a direct connection to the `host:port` specified at connection uri, ignoring the proxy.
This happens on this code from [ManagedChannelImplBuilder](https://github.com/grpc/grpc-java/blob/90aefb26e71fcc8d3f45fa5580dd3131b262db27/core/src/main/java/io/grpc/internal/ManagedChannelImplBuilder.java#L346)
```
  public ManagedChannelImplBuilder(SocketAddress directServerAddress, String authority,
      @Nullable ChannelCredentials channelCreds, @Nullable CallCredentials callCreds,
      ClientTransportFactoryBuilder clientTransportFactoryBuilder,
      @Nullable ChannelBuilderDefaultPortProvider channelBuilderDefaultPortProvider) {
    this.target = makeTargetStringForDirectAddress(directServerAddress);
    this.channelCredentials = channelCreds;
    this.callCredentials = callCreds;
```

I am not sure if there is a special reason for not opening the connection with the host and port values, instead of the `java.net.SocketAddress`:
```
builder = NettyChannelBuilder.forAddress(
                    location.getUri().getHost(), location.getUri().getPort());
```
But opening the channel in the previous way makes GRPC to follow the specified proxy at JVM properties.
`-Dhttps.proxyHost=<host> -Dhttps.proxyPort=<port>`

Could this be a valid enhancement request?

ennuite:
### What's Changed
The Flight SQL JDBC driver ignored JVM proxy settings (`-Dhttps.proxyHost`, `-Dhttps.proxyPort`). Connections would always go directly to the target host, bypassing any configured proxy.

Switched to `NettyChannelBuilder.forAddress(host, port)` for the TCP-based schemes. This causes gRPC to go through `ProxySelector`, which picks up the standard JVM proxy properties.

### Are these changes tested?
Yes. 

Added a test to `ConnectionTest` that installs a recording `ProxySelector` as the JVM default, opens a JDBC connection, and asserts that `ProxySelector.select() `was called. This directly validates that the driver participates in JVM proxy detection without requiring a real proxy server.

Closes #577.

### DIFF
--- a/flight/flight-core/src/main/java/org/apache/arrow/flight/grpc/NettyClientBuilder.java
+++ b/flight/flight-core/src/main/java/org/apache/arrow/flight/grpc/NettyClientBuilder.java
@@ -139,7 +139,9 @@ public class NettyClientBuilder {
       case LocationSchemes.GRPC_INSECURE:
       case LocationSchemes.GRPC_TLS:
         {
-          builder = NettyChannelBuilder.forAddress(location.toSocketAddress());
+          builder =
+              NettyChannelBuilder.forAddress(
+                  location.getUri().getHost(), location.getUri().getPort());
           break;
         }
       case LocationSchemes.GRPC_DOMAIN_SOCKET:
--- a/flight/flight-core/src/main/java/org/apache/arrow/flight/grpc/NettyClientBuilder.java
+++ b/flight/flight-core/src/main/java/org/apache/arrow/flight/grpc/NettyClientBuilder.java
@@ -140,9 +140,9 @@ public class NettyClientBuilder {
       case LocationSchemes.GRPC_TLS:
         {
           final int port = location.getUri().getPort();
-          if (port < 1 || port > 65535) {
+          if (port < 0 || port > 65535) {
             throw new IllegalArgumentException(
-                "Invalid port " + port + ": must be between 1 and 65535.");
+                "Invalid port " + port + ": must be between 0 and 65535.");
           }
           builder = NettyChannelBuilder.forAddress(location.getUri().getHost(), port);
           break;
--- a/flight/flight-core/src/main/java/org/apache/arrow/flight/grpc/NettyClientBuilder.java
+++ b/flight/flight-core/src/main/java/org/apache/arrow/flight/grpc/NettyClientBuilder.java
@@ -139,9 +139,12 @@ public class NettyClientBuilder {
       case LocationSchemes.GRPC_INSECURE:
       case LocationSchemes.GRPC_TLS:
         {
-          builder =
-              NettyChannelBuilder.forAddress(
-                  location.getUri().getHost(), location.getUri().getPort());
+          final int port = location.getUri().getPort();
+          if (port < 1 || port > 65535) {
+            throw new IllegalArgumentException(
+                "Invalid port " + port + ": must be between 1 and 65535.");
+          }
+          builder = NettyChannelBuilder.forAddress(location.getUri().getHost(), port);
           break;
         }
       case LocationSchemes.GRPC_DOMAIN_SOCKET:
--- a/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/utils/ArrowFlightConnectionConfigImpl.java
+++ b/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/utils/ArrowFlightConnectionConfigImpl.java
@@ -55,12 +55,8 @@ public final class ArrowFlightConnectionConfigImpl extends ConnectionConfigImpl 
    *
    * @return the port.
    */
-  public int getPort() throws SQLException {
-    final int port = ArrowFlightConnectionProperty.PORT.getInteger(properties);
-    if (port < 1 || port > 65535) {
-      throw new SQLException("Invalid port " + port + ": must be between 1 and 65535.");
-    }
-    return port;
+  public int getPort() {
+    return ArrowFlightConnectionProperty.PORT.getInteger(properties);
   }
 
   /**
--- a/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/utils/ArrowFlightConnectionConfigImpl.java
+++ b/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/utils/ArrowFlightConnectionConfigImpl.java
@@ -55,8 +55,12 @@ public final class ArrowFlightConnectionConfigImpl extends ConnectionConfigImpl 
    *
    * @return the port.
    */
-  public int getPort() {
-    return ArrowFlightConnectionProperty.PORT.getInteger(properties);
+  public int getPort() throws SQLException {
+    final int port = ArrowFlightConnectionProperty.PORT.getInteger(properties);
+    if (port < 1 || port > 65535) {
+      throw new SQLException("Invalid port " + port + ": must be between 1 and 65535.");
+    }
+    return port;
   }
 
   /**
--- a/flight/flight-sql-jdbc-core/src/test/java/org/apache/arrow/driver/jdbc/ConnectionTest.java
+++ b/flight/flight-sql-jdbc-core/src/test/java/org/apache/arrow/driver/jdbc/ConnectionTest.java
@@ -768,4 +768,118 @@ public class ConnectionTest {
       assertTrue(resultSets[i].isClosed());
     }
   }
+
+  /**
+   * Test that JDBC driver respects JVM proxy settings.
+   *
+   * <p>This test verifies that when JVM proxy properties are set, the Flight client attempts to
+   * connect through the proxy. This will FAIL with the current implementation and PASS when the fix
+   * is applied.
+   *
+   * @throws Exception on error.
+   */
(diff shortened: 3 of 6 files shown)


jbonofre:
As I'm close to release Arrow Java 19.0.0, I will consider this PR for 20.0.0.

ennuite:
@jbonofre yeah that makes sense, this one is still a draft as I have not finished work on it, I will ping you once I finish the PR. Should be ready in the next couple of days.

ennuite:
@jbonofre Do you think this bug merits the introduction of a test? In general I don't like submitting PRs without tests, but in this case I t

It doesn't make much sense to introduce integration tests with proxies, and having a unit test that sets the proxy and checks that the driver points towards it doesn't seem very useful to the codebase as a whole.

If you have an idea for a better test (or if you disagree) let me know, and I'll implement one.

ennuite:
While the fix was in the `NettyClientBuilder`, the bug report concerned the JDBC driver. Because of that, I targeted the test at the driver, ensuring that the whole connection creation flow did not bypass proxies.

Let me know if it's worth it to expand the testing surface with a unit test closer to the place of the bug fix.

lidavidm:
Can we rebase? It seems CI did not trigger

lidavidm:
It seems all workflows time out here, even after a retry.

ennuite:
@lidavidm Thanks for the ping. I rebased all PRs.

ennuite:
@lidavidm 

> It seems all workflows time out here, even after a retry.

Do you guys need help with the CI? I was going to look into some CI stuff on Arrow Go, but given that we have a release here I can jump in before I move to that.

Edit- I synced with JB and I saw that he is already on it. I'll continue with the protocol change implementation in other languages to have feature parity.

lidavidm:
Well, other PRs are passing fine. Perhaps something in this PR may be related?

jbonofre:
@ennuite @lidavidm I'm doing a new review on this one.

ennuite:
@lidavidm you are right. @jbonofre the problem is in my code, it changes port validation and has the side effect of a bug in `testUnencryptedConnectionProvidingInvalidPort`.

I'm looking to fix this now.

ennuite:
I tracked this down all the way to grpc-java and IMO the root cause is there, my change in this PR surfaced the problem. 

The situation looks a bit involved. I can create some PRs in Arrow Java to gain some robustness over these lower level errors in time for this release, or we can leave this PR for a later release. I have no strong opinion either way, but I do have bandwidth to tackle at least one of these right now. If you are interested @lidavidm and @jbonofre let me know and I'll start work on this.

**Improvements we could have in the JDBC driver**:
- Port validation in the JDBC driver in the path of [DriverManager.getConnection()](https://docs.oracle.com/en/java/javase/17/docs/api/java.sql/java/sql/DriverManager.html#getConnection(java.lang.String)): we could validate that the port is numeric and in the range [1, 65535]. This would fix the problem in the test and allow this PR to be merged. It is also quite easy to do.
- Liveness hardening on JDBC connection establishment: could honor the limits of [DriverManager.setLoginTimeout(int)](https://docs.oracle.com/en/java/javase/17/docs/api/java.sql/java/sql/DriverManager.html#setLoginTimeout(int)) and properly implement  [DataSource.setLoginTimeout(int)](https://docs.oracle.com/en/java/javase/17/docs/api/java.sql/javax/sql/DataSource.html#setLoginTimeout(int)) and [DataSource.getLoginTimeout()](https://docs.oracle.com/en/java/javase/17/docs/api/java.sql/javax/sql/DataSource.html#getLoginTimeout()). This would not fix the bug in this test, but would allow users of the driver to have a way to prevent being blocked in the future by bugs that cause connection establishment to hang indefinitely.



**Details about the root cause** (no need to read this unless you're interested in the gRPC internals):

1- I changed the NettyChannelBuilder call from [forAddress(SocketAddress)](https://grpc.github.io/grpc-java/javadoc/io/grpc/netty/NettyChannelBuilder.html#forAddress(java.net.SocketAddress)) to [forAddress(String, int)](https://grpc.github.io/grpc-java/javadoc/io/grpc/netty/NettyChannelBuilder.html#forAddress(java.lang.String,int)). This allows us to respect JVM proxy settings and is the overload gRPC recommends for normal TCP channels.
2-An unintended side effect is that port validation moved from synchronous `SocketAddress` construction to asynchronous DNS resolution in a separate thread owned by gRPC.
3-If that asynchronous validation fails because of an invalid port, gRPC does not propagate the exception through the resolver listener. This causes `testUnencryptedConnectionProvidingInvalidPort` to hang indefinitely and CI to time out.

I will take this up with the gRPC folks and see if they agree that we need to propagate the exception there. In any case that will be too slow for this release.

lidavidm:
Thanks for digging into it. I think we should do the validation here, and implement the timeout and follow up with grpc-java separately.

ennuite:
@lidavidm I'll add the port validation in this PR later today. I think that makes sense as it both unblocks the PR and it is my change here that removed the fail-fast validation.

I created a couple issues for the JDBC connection timeouts: https://github.com/apache/arrow-java/issues/1267 https://github.com/apache/arrow-java/issues/1268 I'll grab them once I finish here, I already familiarized myself with the relevant code yesterday. I think it's fine to leave them for a later release though.

I'll also follow-up with gRPC (and try to provide a PR there if they agree that the problem is on their side).

Thanks for the guidance!

Edit- I will finish this PR tomorrow, as I had a small health hiccup and can't look at screens for a long time for the rest of today.

ennuite:
Validation done, all tests pass locally.

For the validation, I allowed port 0 to be used for compatibility with a few existing tests that used it. I don't think we should allow it though: this is a client code path, it is meaningless to try to connect to port 0.

If you agree with this assessment @lidavidm I'll go ahead and create a separate issue+PR for removing 0 from the valid range (we can leave it for a next release).

jbonofre:
@ennuite I'm doing a review as well. By the way, we pushed a fix on `main` regarding the CI, a rebase would be welcome.

ennuite:
@jbonofre thank you for reviewing! I rebased, let me know of anything else you need and if you have a different opinion on anything that is being discussed.

jbonofre:
@ennuite we will probably need to rebase again once #1270 is merged (I'm on it).

lidavidm:
> Validation done, all tests pass locally.
> 
> For the validation, I allowed port 0 to be used for compatibility with a few existing tests that used it. I don't think we should allow it though: this is a client code path, it is meaningless to try to connect to port 0.
> 
> If you agree with this assessment @lidavidm I'll go ahead and create a separate issue+PR for removing 0 from the valid range (we can leave it for a next release).

Sounds good to me.

ennuite:
I saw that https://github.com/apache/arrow-java/pull/1270 just got merged, I rebased this PR.

flight/flight-core/src/main/java/org/apache/arrow/flight/grpc/NettyClientBuilder.java
```diff
@@ -139,7 +139,12 @@ public NettyChannelBuilder build() {
       case LocationSchemes.GRPC_INSECURE:
       case LocationSchemes.GRPC_TLS:
         {
-          builder = NettyChannelBuilder.forAddress(location.toSocketAddress());
+          final int port = location.getUri().getPort();
+          if (port < 0 || port > 65535) {
+            throw new IllegalArgumentException(
```
lidavidm, line 144: Just to confirm, this is the same exception class that gRPC used to throw for us?

ennuite: Yes, but it wasn't gRPC that used to throw for us.

Details:
`location.toSocketAddress())` is at https://github.com/apache/arrow-java/blob/4e8faf74a3081663b68634d5ab33d81c52c46319/flight/flight-core/src/main/java/org/apache/arrow/flight/Location.java#L67 

The call to [`InetSocketAddress(String hostname, int port)`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/net/InetSocketAddress.html#%3Cinit%3E(java.lang.String,int)) throws `IllegalArgumentException` when the port is smaller than 0 or greater than 65535 (you can find that in the linked Javadoc).

lidavidm (approved):

PR #1062, opened 2026-03-09
GH-1061: Add codegen classifier jar for arrow-vector.

### LINKED ISSUE #1061: Add new classifier jar for arrow-vector tdd files

We have a need for the arrow-vector .tdd files during code generation. These were removed in https://github.com/apache/arrow/commit/01d2fa0d461869a07b2ffeee517beb8116bd0ce2 since there was no know usage of them.

This proposal is to add a new arrow-vector codegen jar file that contains the tdd files.

lriggs:
## What's Changed

Add a new codegen classifier jar for arrow-vector that contains tdd and other template files.

Closes #1061 .

### DIFF
--- a/vector/pom.xml
+++ b/vector/pom.xml
@@ -209,7 +209,7 @@ under the License.
               (comparisons, casts, aggregations) benefit from generating type-specialized code rather
               than hand-writing implementations for each type.
               The TDD files provide a machine-readable definition of Arrow's types that enables this.
-              Including them in the distribution allows downstream projects to generate code that 
+              Including them in the distribution allows downstream projects to generate code that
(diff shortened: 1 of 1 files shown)


telemenar:
> Can you explain how these are useful?

Arrow's type system has many types, and projects building operations across Arrow data (comparisons, casts, aggregations) benefit from generating type-specialized code rather than hand-writing implementations for each type.

The TDD files provide a machine-readable definition of Arrow's types that enables this. Including them in the distribution allows downstream projects to generate code that stays in sync as Arrow's type system evolves.

lriggs:
I added a comment about the usefulness of the codegen files.

PR #1064, opened 2026-03-10
GH-1063: Add is_update field to ActionCreatePreparedStatementResult

### LINKED ISSUE #1063: JDBC driver - Add boolean field to ActionCreatePreparedStatementResult to indicate whether execution path should use CommandPreparedStatementQuery or CommandPreparedStatementUpdate

Following the [discussion in the mailing list](https://lists.apache.org/list.html?dev@arrow.apache.org), the driver should decide the proper RPC to execute a prepared statement based on the new field `is_update` of the `ActionCreatePreparedStatementResult`

ennuite:
## What's Changed

A new field,   `optional bool is_update = 4;`, was added to `message ActionCreatePreparedStatementResult`. When this field is sent by the server, its value indicates whether the proper network flow to execute the query that the driver should follow uses `CommandPreparedStatementQuery` or `CommandPreparedStatementUpdate`.

For outdated servers that don't send the field, the driver maintains its current behavior of using `CommandPreparedStatementQuery` when the `dataset_schema` is not empty, thus ensuring the backward compatibility of the new driver with old servers.

This change was created with AI assistance (Augment Code and Claude code). All lines were manually reviewed by a human. The output is not copyrightable subject matter.

- Closes #1063

### DIFF
--- a/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/client/ArrowFlightSqlClientHandler.java
+++ b/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/client/ArrowFlightSqlClientHandler.java
@@ -388,6 +388,14 @@ public final class ArrowFlightSqlClientHandler implements AutoCloseable {
      */
     Schema getParameterSchema();
 
+    /**
+     * Gets whether this {@link PreparedStatement} is an update statement.
+     *
+     * @return {@code true} if this is an update statement, {@code false} if it's a query, or {@code
+     *     null} if the server did not provide this information.
+     */
+    Boolean isUpdate();
+
     void setParameters(VectorSchemaRoot parameters);
 
     @Override
@@ -479,6 +487,11 @@ public final class ArrowFlightSqlClientHandler implements AutoCloseable {
       @Override
       public void setParameters(VectorSchemaRoot parameters) {
         preparedStatement.setParameters(parameters);
+      }
+
+      @Override
+      public Boolean isUpdate() {
+        return preparedStatement.isUpdate();
       }
 
       @Override
--- a/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/client/ArrowFlightSqlClientHandler.java
+++ b/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/client/ArrowFlightSqlClientHandler.java
@@ -456,6 +456,12 @@ public final class ArrowFlightSqlClientHandler implements AutoCloseable {
 
       @Override
       public StatementType getType() {
+        // If the server provided the is_update field, use it to determine the statement type
+        final Boolean isUpdate = preparedStatement.isUpdate();
+        if (isUpdate != null) {
+          return isUpdate ? StatementType.UPDATE : StatementType.SELECT;
+        }
+        // Fall back to the legacy logic: check if the result set schema is empty
         final Schema schema = preparedStatement.getResultSetSchema();
         return schema.getFields().isEmpty() ? StatementType.UPDATE : StatementType.SELECT;
       }
--- a/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/client/ArrowFlightSqlClientHandler.java
+++ b/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/client/ArrowFlightSqlClientHandler.java
@@ -394,7 +394,7 @@ public final class ArrowFlightSqlClientHandler implements AutoCloseable {
      * @return {@code true} if this is an update statement, {@code false} if it's a query, or {@code
      *     null} if the server did not provide this information.
      */
-    Boolean isUpdate();
+    @Nullable Boolean isUpdate();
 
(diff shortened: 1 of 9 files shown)


lidavidm:
This should be done on apache/arrow, not apache/arrow-java, which only has a copy of the spec

We will need a client and server implementation, preferably in at least two languages

lidavidm:
What happened to the idea of revising the Execute methods to allow returning either data/row count instead of having to separate the two?

lidavidm:
The community meeting is unfortunately extremely late at night for me.

I am basically encouraging looking at https://github.com/apache/arrow/issues/41840

In particular whether a dedicated RPC for this purpose can be declared, instead of overloading more Flight RPC methods

lidavidm:
@jarohen does this seem reasonable from an XTDB perspective?

Ditto @alamb w.r.t InfluxDB

alamb:
It seems reasonable to me but I don't fully understand the need -- if an application sends an UPDATE it will know i would think (as it would have to send a different schema with the DoPut).

alamb:
Basically my questioni is "when would an actual application not know which command to send"?

ennuite:
Hi @alamb, sorry, I missed your reply until today.

> when would an actual application not know which command to send

The situation arises in API calls that allow executing both DML/DDL and SELECT. There is an extended list in the mailing list, but for PEP-249 drivers it's `cursor.execute()` and for JDBC it's `Statement.execute()` and `PreparedStatement.execute()`. At the driver API level it's not known which kind of query it is.

> if an application sends an UPDATE it will know i would think (as it would have to send a different schema with the DoPut).

To be clear, which `DoPut` are you referring to here? Binding parameters or the one to execute DML/DDL? I believe you're referring to the one to execute DML/DDL (with `CommandPreparedStatementUpdate`). 

The problem is that the application has no way of knowing that this RPC call is the correct one to use. 

After the creation of the prepared statement, the application needs to choose between `DoPut` or `GetFlightInfo`+`DoGet` for query execution. How the application makes this decision is undetermined in Flight SQL, with conflicting implementations across drivers. 

Links to the mailing list:
0- https://lists.apache.org/thread/88msflnwkkw8t81czs2ndqhkn1fb1pxd original discussion thread
1- https://lists.apache.org/thread/53dtvjs1xcg4wgxg938lvqffnj7djss4 spinoff thread of the first one, where @prmoore77 reports that Gizmo users faced this issue and they resorted to having a wrapper on top of the ADBC driver. For JDBC, they have their own fork with extra code to handle the same problem: https://github.com/gizmodata/gizmosql-jdbc-driver/commit/90e9bef6e0a4694155e3d6601a0d08fbecbb66b2#diff-af1bbfdaf2e920478a0225424da0c9115a5a03656c8fa7c525f0fdbe8800ed2eR214
2- https://lists.apache.org/thread/x2l10lsprc2fpzj08zd0vggqvop1zl4z

alamb:
> The problem is that the application has no way of knowing that this RPC call is the correct one to use.

I was thinking that the application would know what SQL it sent (and thus a a DDL or DML) and therefore it *would* know what to do

However, I think I see your point if the "application" is the JDBC driver itself. It would have to introspect into the SQL string to try and guess what type of statement was being sent if the code calling into the driver didn't use a specific API

ennuite:
@alamb 

> It would have to introspect into the SQL string to try

Precisely. I think having the server as the source of truth would be preferable to client-side inspection, hence this proposal.

ennuite:
@alamb do you feel comfortable/interested in providing a vote in the mailing list thread? If not, is there any other PMC that you think I should contact?

This is the vote:
https://lists.apache.org/thread/x2l10lsprc2fpzj08zd0vggqvop1zl4z

flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/client/ArrowFlightSqlClientHandler.java
```diff
@@ -388,6 +388,14 @@ public interface PreparedStatement extends AutoCloseable {
      */
     Schema getParameterSchema();
 
+    /**
+     * Gets whether this {@link PreparedStatement} is an update statement.
+     *
+     * @return {@code true} if this is an update statement, {@code false} if it's a query, or {@code
+     *     null} if the server did not provide this information.
+     */
+    Boolean isUpdate();
```
lidavidm:
```suggestion
    @Nullable Boolean isUpdate();
```

flight/flight-sql-jdbc-core/src/test/java/org/apache/arrow/driver/jdbc/utils/MockFlightSqlProducer.java
lidavidm, line 1: Why do we need the V2? Maybe we can have an overload that takes `@Nullable Boolean` or something to put in the isUpdateMap?

ennuite: You are right, an overload is cleaner naming. Addressed that in the [latest commit](https://github.com/apache/arrow-java/pull/1064/changes/df2b8701c8efbfebeaa14b80c7da6c5a69f5c3e4)

lidavidm (approved):

PR #1081, opened 2026-03-21
GH-838: [FlightSQL][JDBC] Fix timestamp unit conversion and sub-millisecond precision in PreparedStatement parameter binding

### LINKED ISSUE #838: [JAVA] FlightSQL JDBC PreparedStatement#setTimestamp ignores connection timeStampPrecision

The FlightSQL JDBC driver always treats timestamps as milliseconds when sending them to the server, ignoring the connection setting timeStampPrecision (for example microseconds). As a result, when the connection precision is set to microseconds the driver truncates/offsets the timestamp and the value stored on the server is incorrect.

Expected behavior When the JDBC connection option timeStampPrecision is set (e.g. microseconds, milliseconds, nanoseconds), PreparedStatement#setTimestamp (and setTimestamp with timestamptz) should send timestamps with the configured precision so the server receives the correct epoch value.

Actual behavior The driver converts java.sql.Timestamp values to an epoch value assuming milliseconds. If the connection is configured for microseconds (or other precision) the conversion is incorrect and inserted timestamps are wrong.

```
// Connection includes timeStampPrecision parameter (example: microseconds)
String jdbcUrlFlight = "jdbc:arrow-flight-sql://localhost:9994?useEncryption=0&timeStampPrecision=microseconds";
try (Connection conn = DriverManager.getConnection(jdbcUrlFlight);
     PreparedStatement stmt = conn.prepareStatement("INSERT INTO ... VALUES (?, ?, ?, ?, ?)")) {

    for (RowData row : generatedData) {
        stmt.setDate(1, row.c_date());
        stmt.setTime(2, row.c_time());
        stmt.setTime(3, row.c_timetz());
        stmt.setTimestamp(4, row.c_timestamp());      // incorrect conversion happens here
        stmt.setTimestamp(5, row.c_timestamptz());   // and here
        stmt.addBatch();
    }
    stmt.executeBatch();
}
```

Example A — connection configured for microseconds (bug triggered)

- Value I want to insert:
    - Timestamp object: "2024-11-03 12:45:09.869885001"
    - nanos = 869885001
    - fastTime = 1730634309000

- **DateTimeUtils#sqlTimestampToUnixTimestamp(Timestamp, TimeZone) returns: 1730637909869**:
    - Interpreted as milliseconds, that is: **2024-11-03 12:45:09.869** (GMT)    
    
- Final value inserted into server: **1970-01-21 00:43:57.909869**
    - Incorrect because the connection used microsecond precision but driver treated the value as milliseconds.
    
Example B — connection set to milliseconds (works)

- Connection: ?timeStampPrecision=milliseconds
- Value I want to insert:
    - "**2023-11-29 09:47:32.659534860**"
    - nanos = 659534860
    - fastTime = 1701247652000
- Driver sends: 1701251252659 (interpreted as milliseconds)

- Server receives: **2023-11-29 09:47:32.659** — correct when precision is milliseconds.

Environment
  - flight-sql-jdbc-driver-18.3.0.jar
  - Java: 17
  - Server FlightSQL implementation/version: 18.2
  - OS: Windows
  
Relevant code & suspected area

DateTimeUtils#sqlTimestampToUnixTimestamp(Timestamp, TimeZone) appears to be converting timestamps assuming millisecond precision. 
```
public static long sqlTimestampToUnixTimestamp(Timestamp timestamp, TimeZone timeZone) {
        long time = timestamp.getTime();
        LocalDateTime dateTime = timestamp.toLocalDateTime();
        long unixTimestamp = dateTime.toEpochSecond(ZoneOffset.UTC) * 1000L + (long)dateTime.get(ChronoField.MILLI_OF_SECOND);
        if (timeZone != null) {
            unixTimestamp += (long)timeZone.getOffset(time);
        }

        unixTimestamp -= (long)DEFAULT_ZONE.getOffset(time);
        return unixTimestamp;
    }
```

The driver should respect the connection timeStampPrecision parameter, and correctlyreturn the exact epoch value (as a long) at the configured timestamp precision.

dmitry-chirkov-dremio:
### Rationale for this change

The FlightSQL JDBC driver's `PreparedStatement.setTimestamp()` has two bugs (#838):

1. **Unit scaling bug**: `TimestampAvaticaParameterConverter` writes raw epoch millisecond values directly into Arrow timestamp vectors regardless of the vector's time unit. This causes data corruption when the server uses SECOND, MICROSECOND, or NANOSECOND precision — the millisecond value is misinterpreted as the target unit (e.g., epoch millis treated as epoch micros → dates near 1970).

2. **Precision loss bug**: The underlying Avatica library serializes `java.sql.Timestamp` into a `long` (epoch milliseconds) via `TypedValue.ofJdbc()`, discarding the sub-millisecond nanoseconds from `Timestamp.getNanos()` before they reach the Arrow converters. This means even with correct unit scaling, microsecond and nanosecond precision is permanently lost.

**Fix 1 — Unit scaling** (converter layer):
- Modified `TimestampAvaticaParameterConverter` to store the `ArrowType.Timestamp` and scale the epoch millisecond value to the target unit before writing to the vector:
  - SECOND: divide by 1,000 using `Math.floorDiv`
  - MILLISECOND: pass through
  - MICROSECOND: multiply by 1,000
  - NANOSECOND: multiply by 1,000,000

**Fix 2 — Sub-millisecond precision preservation** (JDBC interception layer):
- `ArrowFlightPreparedStatement`: Overrides `setTimestamp()` to capture the raw `java.sql.Timestamp` objects (with full nanosecond precision) before Avatica serializes them to millis. Also overrides `clearParameters()` to clean up.
- `ArrowFlightMetaImpl`: Added `getRawTimestamps()` to bridge the captured timestamps from the statement to the parameter binder.
- `AvaticaParameterBinder`: Updated `bind()` and `BinderVisitor` to propagate the optional raw `Timestamp` to converters.
- `TimestampAvaticaParameterConverter`: Added `bindParameter(vector, typedValue, index, rawTimestamp)` overload and `convertFromTimestamp(Timestamp)` that reconstructs full-precision epoch values from `Timestamp.getTime()` (epoch seconds) + `Timestamp.getNanos()` (fractional second in nanos), avoiding the Avatica millis truncation.
- `ListAvaticaParameterConverter`, `LargeListAvaticaParameterConverter`, `FixedSizeListAvaticaParameterConverter`: Updated to pass `null` as the raw timestamp argument to the new `BinderVisitor` constructor.

**Fix 3 — Stale raw timestamp handling** (overwrite edge cases):
- `ArrowFlightPreparedStatement`: Keeps the side-channel `rawTimestamps` map in sync for `setObject(...)` overwrites by clearing the affected entry.
- `TimestampAvaticaParameterConverter`: Only uses a `rawTimestamp` when the incoming `TypedValue.type` is still `JAVA_SQL_TIMESTAMP`, so stale side-channel entries are ignored if a later setter such as `setLong(...)` overwrites the parameter.

### Are these changes tested?

Yes.
- **Unit tests** (`TimestampAvaticaParameterConverterTest`): cover all Arrow timestamp vector units (with and without timezone) for millis-based scaling, the raw `Timestamp` path for micro/nano precision preservation and milli/second truncation, null-`rawTimestamp` fallback, negative epoch handling, and the stale-map case where a later non-timestamp setter must ignore a stale raw timestamp.
- **Integration tests** (`ArrowFlightPreparedStatementTest`): cover the full `setTimestamp()` → interception → binding → mock server validation path, plus overwrite scenarios including `setObject(...)` after `setTimestamp(...)`, `setLong(...)` after `setTimestamp(...)`, and `setTimestamp(...)` after `setObject(...)`.

**This PR includes breaking changes to public APIs.** `PreparedStatement#setTimestamp()` now correctly converts timestamp values for non-millisecond precision columns. Previously, timestamps inserted via the FlightSQL JDBC driver into SECOND, MICROSECOND, or NANOSECOND columns were silently corrupted. Sub-millisecond precision (microseconds, nanoseconds) is now preserved when the target column supports it.

**This PR contains a "Critical Fix".** The bug caused incorrect data to be produced — epoch milliseconds were written as-is into microsecond/nanosecond vectors, resulting in wildly wrong dates (e.g., 2024 → 1970). Additionally, sub-millisecond precision was silently discarded.

Closes #838

### DIFF
--- a/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightMetaImpl.java
+++ b/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightMetaImpl.java
@@ -19,6 +19,7 @@ package org.apache.arrow.driver.jdbc;
 import java.sql.Connection;
 import java.sql.SQLException;
 import java.sql.SQLTimeoutException;
+import java.sql.Timestamp;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
@@ -31,6 +32,7 @@ import org.apache.arrow.util.Preconditions;
 import org.apache.arrow.vector.types.pojo.Schema;
 import org.apache.calcite.avatica.AvaticaConnection;
 import org.apache.calcite.avatica.AvaticaParameter;
+import org.apache.calcite.avatica.AvaticaStatement;
 import org.apache.calcite.avatica.ColumnMetaData;
 import org.apache.calcite.avatica.MetaImpl;
 import org.apache.calcite.avatica.NoSuchStatementException;
@@ -105,9 +107,10 @@ public class ArrowFlightMetaImpl extends MetaImpl {
       throw new IllegalStateException("Prepared statement not found: " + statementHandle);
     }
 
+    Map<Integer, Timestamp> rawTimestamps = getRawTimestamps(statementHandle);
     new AvaticaParameterBinder(
             preparedStatement, ((ArrowFlightConnection) connection).getBufferAllocator())
-        .bind(typedValues);
+        .bind(typedValues, 0, rawTimestamps);
 
     if (statementHandle.signature == null
         || statementHandle.signature.statementType == StatementType.IS_DML) {
@@ -149,11 +152,12 @@ public class ArrowFlightMetaImpl extends MetaImpl {
       throw new IllegalStateException("Prepared statement not found: " + statementHandle);
     }
 
+    Map<Integer, Timestamp> rawTimestamps = getRawTimestamps(statementHandle);
     final AvaticaParameterBinder binder =
         new AvaticaParameterBinder(
             preparedStatement, ((ArrowFlightConnection) connection).getBufferAllocator());
     for (int i = 0; i < parameterValuesList.size(); i++) {
-      binder.bind(parameterValuesList.get(i), i);
+      binder.bind(parameterValuesList.get(i), i, rawTimestamps);
     }
 
     // Update query
@@ -171,6 +175,14 @@ public class ArrowFlightMetaImpl extends MetaImpl {
      */
     throw AvaticaConnection.HELPER.wrap(
         String.format("%s does not use frames.", this), AvaticaConnection.HELPER.unsupported());
+  }
+
+  private Map<Integer, Timestamp> getRawTimestamps(StatementHandle statementHandle) {
+    AvaticaStatement avaticaStmt = connection.statementMap.get(statementHandle.id);
+    if (avaticaStmt instanceof ArrowFlightPreparedStatement) {
+      return ((ArrowFlightPreparedStatement) avaticaStmt).getRawTimestamps();
+    }
+    return Collections.emptyMap();
   }
 
   private PreparedStatement prepareForHandle(final String query, StatementHandle handle) {
--- a/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightPreparedStatement.java
+++ b/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightPreparedStatement.java
@@ -83,7 +83,7 @@ public class ArrowFlightPreparedStatement extends AvaticaPreparedStatement
   @Override
   public void setTimestamp(int parameterIndex, Timestamp x) throws SQLException {
     if (x != null) {
-      rawTimestamps.put(parameterIndex, x);
+      rawTimestamps.put(parameterIndex, (Timestamp) x.clone());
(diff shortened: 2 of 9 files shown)


jbonofre (approved):
Good one ! Thanks @dmitry-chirkov-dremio !

I just have a suggestion for the `convertFromMillis` method.

xborder (commented):
overall LGTM. I'd probably just add a few tests for edge cases. Examples:
- Change the Timestamp object after setting the parameter. If we're keeping a reference to the object, is it possible to change it after the bind?
- Setting timestamp with null on a string with multiple parameters
- Test setObject
- Use setObject for a timestamp parameter, before/after setTimestamp

vandop:
@dmitry-chirkov-dremio / @xborder are we still blocked on this one? can we get it going?

jbonofre:
@vandop I'm on it.

PR #1099, opened 2026-03-30
GH-1098: [Java][Vector] Fix always-true precondition check in LargeListVector.setValueCount

### LINKED ISSUE #1098: [Java][Vector] Always-true precondition check in LargeListVector.setValueCount allows silent data corruption
### Describe the bug

In `LargeListVector.setValueCount()`, the precondition check that validates `childValueCount` before casting from `long` to `int` uses `||` instead of `&&`, making the condition always evaluate to `true`:

```java
Preconditions.checkArgument(
    childValueCount <= Integer.MAX_VALUE || childValueCount >= Integer.MIN_VALUE,
    "LargeListVector doesn't yet support 64-bit allocations: %s",
    childValueCount);
```

For any `long` value `x`, at least one of `x <= Integer.MAX_VALUE` or `x >= Integer.MIN_VALUE` is always true (their union covers the entire `long` range), so the check never rejects any value.

This means:
- A `childValueCount` exceeding `Integer.MAX_VALUE` silently truncates when cast to `(int)`, causing data corruption
- A negative `childValueCount` (from corrupted offset data) is silently passed through

### Expected behavior

The check should use `&&` with a lower bound of `0` (since `childValueCount` represents element count and must be non-negative):

```java
Preconditions.checkArgument(
    childValueCount <= Integer.MAX_VALUE && childValueCount >= 0,
    "LargeListVector doesn't yet support 64-bit allocations: %s",
    childValueCount);
```

This is consistent with the correct implementation in `LargeListViewVector.setValueCount()` (line 887-890).

### Component(s)

Java

LuciferYang:
## What's Changed

Fix the precondition check in `LargeListVector.setValueCount()` where `||` made the condition always true (`childValueCount <= Integer.MAX_VALUE || childValueCount >= Integer.MIN_VALUE` is a tautology for any `long` value), allowing silent data corruption when `childValueCount` exceeds `Integer.MAX_VALUE` or is negative.

Changes:
- Replace `||` with `&&` and `Integer.MIN_VALUE` with `0` to correctly bound `childValueCount` to `[0, Integer.MAX_VALUE]`, consistent with `LargeListViewVector`
- Add rejection tests for negative and overflow `childValueCount`
- Add positive boundary tests for zero, valid, and `Integer.MAX_VALUE` values

Closes #1098.

### DIFF
--- a/vector/src/main/java/org/apache/arrow/vector/complex/LargeListVector.java
+++ b/vector/src/main/java/org/apache/arrow/vector/complex/LargeListVector.java
@@ -1041,8 +1041,8 @@ public class LargeListVector extends BaseValueVector
      * TODO: revisit when 64-bit vectors are supported
      */
     Preconditions.checkArgument(
-        childValueCount <= Integer.MAX_VALUE && childValueCount >= 0,
-        "LargeListVector doesn't yet support 64-bit allocations: %s",
+        childValueCount >= 0 && childValueCount <= Integer.MAX_VALUE,
+        "LargeListVector childValueCount must be in [0, Integer.MAX_VALUE] but was: %s",
         childValueCount);
     vector.setValueCount((int) childValueCount);
   }
--- a/vector/src/main/java/org/apache/arrow/vector/complex/LargeListVector.java
+++ b/vector/src/main/java/org/apache/arrow/vector/complex/LargeListVector.java
@@ -1041,7 +1041,7 @@ public class LargeListVector extends BaseValueVector
      * TODO: revisit when 64-bit vectors are supported
      */
     Preconditions.checkArgument(
-        childValueCount <= Integer.MAX_VALUE || childValueCount >= Integer.MIN_VALUE,
+        childValueCount <= Integer.MAX_VALUE && childValueCount >= 0,
         "LargeListVector doesn't yet support 64-bit allocations: %s",
         childValueCount);
     vector.setValueCount((int) childValueCount);
--- a/vector/src/test/java/org/apache/arrow/vector/TestLargeListVector.java
+++ b/vector/src/test/java/org/apache/arrow/vector/TestLargeListVector.java
@@ -21,6 +21,7 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertSame;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.util.ArrayList;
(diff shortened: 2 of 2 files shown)


LuciferYang:
Could a maintainer please add the bug-fix label to this PR? I don't have label permissions on this repository. Thank you!

ennuite (approved):
Good one. I looked into this: the bug has been there since the class was created, see https://github.com/apache/arrow-java/commit/4deba4420a1ae90406f89fb18c6529692899607e#diff-c3066ea815144f3e05b277c87cae8a7701f0f1481e15da508261664ab017d944R984
The PR fixes the bug and has test coverage for all equivalence classes. The correct values are in the interval [0, `Integer.MAX_Value`]; there are tests for a negative value, 0, a positive value smaller than `Integer.MAX_Value`, `Integer.MAX_Value` and a positive value greater than `Integer.MAX_Value`.

Thank you

vector/src/main/java/org/apache/arrow/vector/complex/LargeListVector.java
```diff
@@ -1041,7 +1041,7 @@ public void setValueCount(int valueCount) {
      * TODO: revisit when 64-bit vectors are supported
      */
     Preconditions.checkArgument(
-        childValueCount <= Integer.MAX_VALUE || childValueCount >= Integer.MIN_VALUE,
+        childValueCount <= Integer.MAX_VALUE && childValueCount >= 0,
```
axreldable: nit: I'd use `childValueCount >= 0 && childValueCount <= Integer.MAX_VALUE` for the sake of readability.

vector/src/test/java/org/apache/arrow/vector/TestLargeListVector.java
```diff
@@ -1127,4 +1128,90 @@ private void writeIntValues(UnionLargeListWriter writer, int[] values) {
...
+      vector.addOrGetVector(FieldType.nullable(MinorType.INT.getType()));
+      vector.allocateNew();
+
+      // Write a negative value into the offset buffer to simulate a corrupted offset.
+      // The Preconditions check should reject any negative childValueCount.
+      vector.getOffsetBuffer().setLong(LargeListVector.OFFSET_WIDTH, -1L);
+      vector.setLastSet(0);
+
+      assertThrows(
+          IllegalArgumentException.class,
```
axreldable, line 1147: The error message throwing here is a bit confusing: `IllegalArgument LargeListVector doesn't yet support 64-bit allocations: -1`

vector/src/test/java/org/apache/arrow/vector/TestLargeListVector.java
```diff
@@ -1127,4 +1128,90 @@ private void writeIntValues(UnionLargeListWriter writer, int[] values) {
...
+  public void testSetValueCountAcceptsMaxIntChildValueCount() {
+    try (final LargeListVector vector = LargeListVector.empty("list", allocator)) {
+      vector.addOrGetVector(FieldType.nullable(MinorType.INT.getType()));
+      vector.allocateNew();
+
+      // Write Integer.MAX_VALUE into the offset buffer to verify the upper boundary is inclusive.
+      // The Preconditions check should accept this value (not throw IllegalArgumentException).
+      // The child vector may throw OversizedAllocationException due to memory limits,
+      // which is expected and unrelated to the precondition validation.
+      vector.getOffsetBuffer().setLong(LargeListVector.OFFSET_WIDTH, (long) Integer.MAX_VALUE);
```
axreldable: nit: `Casting 'Integer.MAX_VALUE' to 'long' is redundant`

vector/src/test/java/org/apache/arrow/vector/TestLargeListVector.java
```diff
@@ -1127,4 +1128,90 @@ private void writeIntValues(UnionLargeListWriter writer, int[] values) {
...
+        vector.setValueCount(1);
+      } catch (IllegalArgumentException e) {
+        throw new AssertionError(
+            "setValueCount should not reject childValueCount = Integer.MAX_VALUE", e);
+      } catch (Exception e) {
+        // OversizedAllocationException or other allocation errors are expected
+        // when trying to allocate Integer.MAX_VALUE elements — this is fine,
+        // the precondition check itself passed.
+      }
+    }
```
axreldable, line 1217: nit: Maybe this can be replaced with

```
try {
        vector.setValueCount(1);
      } catch (IllegalArgumentException e) {
        fail("setValueCount should not reject childValueCount = Integer.MAX_VALUE", e);
      } catch (OversizedAllocationException | OutOfMemoryException e) {
        // OversizedAllocationException or other allocation errors are expected
        // when trying to allocate Integer.MAX_VALUE elements — this is fine,
        // the precondition check itself passed.
      }
```

axreldable:
Thank you for the bug-fix @LuciferYang ! I've left some comments. My main concern is the confusing error message for small negative values.

LuciferYang:
@axreldable Tbanks for your review.I am on vacation,  I will provide fixes and feedback on your suggestions in 3 days

vector/src/test/java/org/apache/arrow/vector/TestLargeListVector.java
LuciferYang: [1717f2e](https://github.com/apache/arrow-java/pull/1099/commits/1717f2e77e9f37490aebee5b730a7c2dd082495a) address the comments

axreldable (approved):

lidavidm:
Sorry @LuciferYang can you rebase again to check CI?

LuciferYang:
rebased

lidavidm (approved):

LuciferYang:
Thank you @lidavidm @axreldable @ennuite

PR #1101, opened 2026-03-31
revert: downgrade Derby 10.17.1.0 → 10.15.2.0 (Java 19+ incompatibility)

mateusaubin:
## Summary

- Reverts dependabot bump b8b597a0 that upgraded Derby from `10.15.2.0` to `10.17.1.0`
- Derby 10.17.1.0 requires Java 19+ due to use of restricted `sun.misc.Unsafe` APIs — breaks builds on our current JDK target
- No backport has been released by Derby maintainers (fix exists in repo but was never shipped); Derby is effectively unmaintained

## Context

Derby is a **test-scope only** dependency in `flight/flight-sql` — it is not shipped to users. This was discussed in the daily sync and needs to be annotated in Vanta accordingly.

## Options considered

| Option | Notes |
|--------|-------|
| ✅ **Revert (this PR)** | Unblocks builds immediately; accept risk given test-scope |
| Remove Derby + disable tests | Cleaner long-term if we don't need it |
| Replace with H2 | More actively maintained embedded DB |
| Manage our own Derby `10.16.1.2` release | Full control but high maintenance burden |
| Upgrade arrow-java/JDBC to JDK-21 | Would allow keeping `10.17.1.0`; larger scope change |

## Follow-up

- [ ] Annotate in Vanta: test-scope dependency, originates from upstream arrow-java fork, not shipped to users
- [ ] Decide long-term Derby strategy (see options above)

## Test plan

- [ ] Build passes on current JDK target after revert
- [ ] `flight/flight-sql` tests pass with Derby `10.15.2.0`

### DIFF
--- a/adapter/avro/pom.xml
+++ b/adapter/avro/pom.xml
@@ -23,7 +23,7 @@ under the License.
   <parent>
     <groupId>org.apache.arrow</groupId>
     <artifactId>arrow-java-root</artifactId>
-    <version>19.0.0-SNAPSHOT</version>
+    <version>19.0.0-iomete.1</version>
     <relativePath>../../pom.xml</relativePath>
   </parent>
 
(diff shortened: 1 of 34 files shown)


mateusaubin:
Opened by mistake against upstream; re-opening against iomete/arrow-java fork.

Issue #1116, opened 2026-04-15
[IPC][ZSTD] Compressed buffer prefix can be written as 0 while ZSTD frame has non-zero content size


m8928:
### Describe the bug

We observed Arrow IPC streams written by Arrow Java where a ZSTD-compressed buffer has an invalid 8-byte uncompressed-length prefix.

The compressed payload itself appears to be a valid ZSTD frame, but the Arrow IPC prefix is `0`, while the ZSTD frame reports a non-zero decompressed content size.

This makes downstream readers fail. In our case ClickHouse 26.1 fails with:

```text
Error while reading batch of Arrow data:
IOError: ZSTD decompression failed: Destination buffer is too small
```

The same file also fails when loaded with Arrow Java `ArrowStreamReader`.

This has been observed with both:

- Arrow Java `18.3.0` + `zstd-jni 1.5.7-6`
- Arrow Java `19.0.0` + `zstd-jni 1.5.7-7`

### Environment

Observed with both dependency sets:

1. Original environment
   - Apache Arrow Java: `18.3.0`
   - `org.apache.arrow:arrow-vector:18.3.0`
   - `org.apache.arrow:arrow-memory-netty:18.3.0`
   - `org.apache.arrow:arrow-compression:18.3.0`
   - `com.github.luben:zstd-jni:1.5.7-6`

2. Also reproduced with
   - Apache Arrow Java: `19.0.0`
   - `org.apache.arrow:arrow-vector:19.0.0`
   - `org.apache.arrow:arrow-memory-netty:19.0.0`
   - `org.apache.arrow:arrow-compression:19.0.0`
   - `com.github.luben:zstd-jni:1.5.7-7`

Other context:

- Java: 17
- Writer API:
  - `ArrowStreamWriter`
  - `CommonsCompressionFactory.INSTANCE`
  - `CompressionUtil.CodecType.ZSTD`

Writer creation:

```java
new ArrowStreamWriter(
    root,
    null,
    channel,
    IpcOption.DEFAULT,
    CommonsCompressionFactory.INSTANCE,
    CompressionUtil.CodecType.ZSTD
);
```

### What happened

One generated Arrow IPC stream contains 44 record batches and a valid end marker. The total row count from record batch metadata matches the sidecar count file.

However, batch 21 fails to load.

Inspection results:

```text
file size: 26090728 bytes
schema fields: 504
record batches: 44
batch metadata row sum: 92936
sidecar count: 92936
```

The failure occurs on batch 21.

[...]

PR #1119, opened 2026-04-18
GH-1116: [Java] Fix compressed buffer prefix write and ZSTD dstCapacity

LuciferYang:
## What

Two fixes in the compression codec:

1. **`AbstractCompressionCodec.compress()`: capture `writerIndex()` once**

   The previous code read `uncompressedBuffer.writerIndex()` at multiple sites — for the size comparison and again after `doCompress()` to populate the 8-byte uncompressed-length prefix. Capture the value once at the top of `compress()` and reuse it for the empty-buffer check, the size comparison, and the prefix, so all three consumers see the same value.

2. **`ZstdCompressionCodec.doCompress()`: `dstCapacity` overstated by 8 bytes**

   `Zstd.compressUnsafe(dst, dstSize, ...)` expects `dstSize` to be the available space from `dst`. The code offsets `dst` by 8 bytes past the prefix but passed `8 + maxSize` instead of `maxSize`. The `compressBound()` headroom hides this in practice, but the parameter was semantically wrong. Pass `maxSize`.

## Tests

Covered by the existing round-trip tests (`testEmptyBuffer`, `testReadWriteStream`, `testReadWriteFile`, etc.). I was not able to construct a minimal reproducer for the original `declaredUncompressed=0` symptom on the unfixed code, so both fixes are conservative correctness improvements derived from code inspection rather than failing-then-green regression tests.

### DIFF
--- a/vector/src/main/java/org/apache/arrow/vector/compression/AbstractCompressionCodec.java
+++ b/vector/src/main/java/org/apache/arrow/vector/compression/AbstractCompressionCodec.java
@@ -29,7 +29,14 @@ public abstract class AbstractCompressionCodec implements CompressionCodec {
 
   @Override
   public ArrowBuf compress(BufferAllocator allocator, ArrowBuf uncompressedBuffer) {
-    if (uncompressedBuffer.writerIndex() == 0L) {
+    // Capture the uncompressed length once upfront to avoid any inconsistency from
+    // re-reading writerIndex() at different points. Since the uncompressedBuffer may be
+    // a shared reference to a vector's internal buffer, reading writerIndex() only once
+    // ensures the same value is used for the empty-buffer check, compression, size
+    // comparison, and the 8-byte uncompressed-length prefix.
+    long uncompressedLength = uncompressedBuffer.writerIndex();
+
+    if (uncompressedLength == 0L) {
       // shortcut for empty buffer
       ArrowBuf compressedBuffer = allocator.buffer(CompressionUtil.SIZE_OF_UNCOMPRESSED_LENGTH);
       compressedBuffer.setLong(0, 0);
@@ -41,7 +48,6 @@ public abstract class AbstractCompressionCodec implements CompressionCodec {
     ArrowBuf compressedBuffer = doCompress(allocator, uncompressedBuffer);
     long compressedLength =
         compressedBuffer.writerIndex() - CompressionUtil.SIZE_OF_UNCOMPRESSED_LENGTH;
-    long uncompressedLength = uncompressedBuffer.writerIndex();
 
     if (compressedLength > uncompressedLength) {
       // compressed buffer is larger, send the raw buffer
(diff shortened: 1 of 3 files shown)


vector/src/main/java/org/apache/arrow/vector/compression/AbstractCompressionCodec.java
lidavidm, line 1: While I think this change is fine, nothing should be modifying the buffer during this method. I don't believe any vectors/etc. are safe to use concurrently, so the comment is misleading about the guarantees we can provide.

LuciferYang: Good point — the original framing implied a concurrency guarantee we don't actually provide, and there's no in-tree `doCompress` implementation that mutates the source buffer's `writerIndex`. Tightened the comment to state only the engineering invariant (three in-method consumers must observe the same value) and dropped the "shared reference" / `doCompress` speculation:

```java
// GH-1116: capture writerIndex() once so the empty-buffer check, size
// comparison, and uncompressed-length prefix all see the same value.
long uncompressedLength = uncompressedBuffer.writerIndex();
```

Pushed in d7b6a8fb5.

lidavidm:
@LuciferYang were you able to replicate the original issue, by the way?

LuciferYang:
Honest answer: no. I built the two new tests around the shape of data the reporter described (all-null `TimestampMilli`, 3469 rows, `VectorSchemaRoot` reuse via `clear()`/`allocateNew()`), but I just re-checked with the source change reverted — both still pass. So they're round-trip coverage for that shape, not a direct regression test for the `declaredUncompressed=0` symptom.

The fix came from reading the code: the old `compress()` reads `writerIndex()` twice — once for the size comparison, once for the 8-byte prefix after `doCompress()`. Reading once and reusing the value is the conservative invariant. The ZSTD `dstCapacity` change is similar — the `compressBound()` headroom hides it, but the parameter was semantically wrong.

If you'd rather hold until we have a failing-then-green test, that's fair. Otherwise happy to dig further — the reporter's env was Arrow 18.3.0/19.0.0 with zstd-jni 1.5.7-6/-7, but the allocator setup wasn't in the report.

lidavidm:
I think we should document that the new tests don't actually reflect the original bug report, at least (or consider if they're necessary even). But the fixes themselves are worth having.

LuciferYang:
Fair. Dropped both tests in b8e6b46 — `testEmptyBuffer` and `testReadWriteStream` already cover those round-trip shapes.

lidavidm (approved):
Thanks!

LuciferYang:
Thanks for your review @lidavidm

Dan-Nedelescu:
Hi @LuciferYang thanks for this fix! I'm interested in using this fix for my project. Do you know what the release schedule might be here?

LuciferYang:
Sorry, I don't know. I am not a PMC Member of  Apache Arrow

Dan-Nedelescu:
No worries and thanks for the response!

PR #1120, opened 2026-04-18
GH-1117: Use updated handle in prepared stmt

### LINKED ISSUE #1117: [Flight SQL] ClosePreparedStatement uses original handle instead of updated handle from server

When using the `FlightSqlClient.PreparedStatement` flow, the client uses the original prepared statement handle when invoking `ClosePreparedStatement`, rather than the updated handle returned by the server when binding parameters.

## Observed behavior

- After preparing a statement, the server may return an updated handle in binding parameters call.
- The client correctly uses the updated handle for `Execute`.
- However, when calling `ClosePreparedStatement`, the client still uses the original handle stored in the final field `preparedStatementResult`.

## Expected behavior

Based on my understanding of the protocol, once the server returns an updated handle, the client should consistently use that handle for all subsequent operations.

By specification on FlightSql.proto file:

```
  // Because the handle could potentially be updated, any previous handles for this prepared
  // statement should be considered invalid, and all subsequent requests for this prepared
  // statement must use this new handle.
```

## Impact

In a custom server implementation I'm developing, this causes inconsistencies because:

- the server expects the updated handle for all follow-up calls, including new put parameter calls and ClosePreparedStatement
- `ClosePreparedStatement` fails or behaves unexpectedly when the original handle is used

## Question / clarification

Is this the intended behavior according to the Flight SQL protocol, or is this a bug in the client implementation?

If this is expected, It's quite simple to adjust my server logic accordingly, but wanted to confirm that is not a bug on client implementation.

ennuite:
## What's Changed
When a server sent an updated handle for a prepared statement after parameters were bound,
the client would use the new handle for the subsequent call to `GetFlightInfo` and then discard it. Any other
future requests for that prepared statement would use the original handle.

With these changes, the updated handle is stored and reused for future requests.

Closes #1117.

### DIFF
--- a/flight/flight-sql/src/main/java/org/apache/arrow/flight/sql/FlightSqlClient.java
+++ b/flight/flight-sql/src/main/java/org/apache/arrow/flight/sql/FlightSqlClient.java
@@ -1217,6 +1217,7 @@ public class FlightSqlClient implements AutoCloseable {
   public static class PreparedStatement implements AutoCloseable {
     private final FlightClient client;
     private final ActionCreatePreparedStatementResult preparedStatementResult;
+    private ByteString handle;
     private VectorSchemaRoot parameterBindingRoot;
     private boolean isClosed;
     private Schema resultSetSchema;
@@ -1229,6 +1230,7 @@ public class FlightSqlClient implements AutoCloseable {
       preparedStatementResult =
           FlightSqlUtils.unpackAndParseOrThrow(
               preparedStatementResults.next().getBody(), ActionCreatePreparedStatementResult.class);
+      handle = preparedStatementResult.getPreparedStatementHandle();
       isClosed = false;
     }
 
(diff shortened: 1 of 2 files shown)


xborder (approved):
LGTM. I think you could expand the tests to other scenarios (e.g ensure the handle is the same during `ActionClosePreparedStatement` after `DoGet`)

ennuite:
> LGTM. I think you could expand the tests to other scenarios (e.g ensure the handle is the same during `ActionClosePreparedStatement` after `DoGet`)

The test does verify this case.

xborder:
I've talked with @ennuite. My suggestion is to add another test that does `Create->GetFlightInfo->DoGet->Close` and checks the handle didn't change. Just to expand the testing surface

jbonofre:
@ennuite @xborder I fixed the CI on `main`. Can you please rebase?

ennuite:
@xborder I added the test. @jbonofre I rebased, can I get a bug-fix label?

lidavidm:
@ennuite do you mind rebasing again?

lidavidm (approved):

PR #1136, opened 2026-05-06
GH-1125: Read empty ListVector correctly using UnionListReader after IPC deser

### LINKED ISSUE #1125: `UnionListReader.setPosition` throws IOOBE on a post-IPC empty List

On 19.0.0 (and on master), calling `setPosition` on a `ListVector` that came off the wire via IPC with `valueCount == 0` throws `IndexOutOfBoundsException`.

Some callers may hit this without touching `setPosition` themselves as
- `SingleStructReaderImpl.reader(String)` calls `setPosition(idx())` the first time a child reader on a struct is looked up.
- `UnionMapReader` inherits from `UnionListReader` and breaks the same way.
- `UnionLargeListReader` has a related variant; capacity guard seems to be missing and it fails whenever the offset buffer is shorter than `(idx + 2) * 8` bytes.

The reader code hasn't changed. `ListVector.setReaderAndWriterIndex` changes and now correctly emits the leading `[0]` offset when `valueCount == 0` (as per the Arrow spec), so the receiver's offset buffer is no longer "zero-capacity" and if block `capacity() == 0` never enters.

### Repro

The shape we actually hit — a struct with a list-typed child, struct itself having `valueCount == 0`:

```java
StructVector parent = ...;            // child "l" is LIST<INT>, valueCount == 0
parent.getReader().reader("l");       // IOOBE
```

bodduv:
Arrow IPC represents variable-length vectors with an offset buffer containing `valueCount + 1` offsets. For an empty `ListVector`, that still means the serialized and deserialized vector can have a non-empty offset buffer containing the leading zero offset. This is correct according to the Arrow layout, but it exposes a bug at `UnionListReader.setPosition` and other similar places. `UnionListReader.setPosition(0)` used offset-buffer capacity as the empty-vector check. That worked only when the offset buffer had zero capacity. After IPC, the empty vector has non-zero offset-buffer capacity, so the reader could throw `IndexOutOfBoundsException`. `UnionLargeListReader` has the same logical issue and also lacked the empty-buffer guard.

## What's Changed

Update `UnionListReader` and `UnionLargeListReader` to validate reader positioning against `valueCount` instead of treating offset-buffer capacity as the logical row boundary. All out-of-range positions will throw. For valid non-empty positions, the readers also defensively verify that the offset buffer has enough capacity for both `index` and `index + 1` before reading offsets.

The shared bounds logic is kept in a package-private `UnionListReaderBoundsChecker` helper so `UnionListReader` and `UnionLargeListReader` use the reuse code.

Closes #1125.

### DIFF
--- a/vector/src/main/java/org/apache/arrow/vector/complex/impl/UnionLargeListReader.java
+++ b/vector/src/main/java/org/apache/arrow/vector/complex/impl/UnionLargeListReader.java
@@ -53,9 +53,24 @@ public class UnionLargeListReader extends AbstractFieldReader {
 
   @Override
   public void setPosition(int index) {
+    int valueCount = vector.getValueCount();
+    if (UnionListReaderBoundsChecker.isEmptyVectorPosition(index, valueCount)) {
+      setEmptyPosition(index);
+      return;
+    }
+
+    UnionListReaderBoundsChecker.checkIndex(index, valueCount);
+    UnionListReaderBoundsChecker.checkOffsetBuffer(vector.getOffsetBuffer(), index, OFFSET_WIDTH);
+
     super.setPosition(index);
-    currentOffset = vector.getOffsetBuffer().getLong((long) index * OFFSET_WIDTH) - 1;
-    maxOffset = vector.getOffsetBuffer().getLong(((long) index + 1L) * OFFSET_WIDTH);
+    currentOffset = vector.getElementStartIndex(index) - 1;
+    maxOffset = vector.getElementEndIndex(index);
+  }
+
+  private void setEmptyPosition(int index) {
+    super.setPosition(index);
+    currentOffset = 0;
+    maxOffset = 0;
   }
 
   @Override
--- a/vector/src/main/java/org/apache/arrow/vector/complex/impl/UnionLargeListReader.java
+++ b/vector/src/main/java/org/apache/arrow/vector/complex/impl/UnionLargeListReader.java
@@ -54,13 +54,14 @@ public class UnionLargeListReader extends AbstractFieldReader {
   @Override
   public void setPosition(int index) {
     int valueCount = vector.getValueCount();
-    if (UnionListReaderBoundsChecker.isEmptyVectorPosition(index, valueCount)) {
+    if (valueCount == 0 && index == 0) {
       setEmptyPosition(index);
       return;
     }
 
-    UnionListReaderBoundsChecker.checkIndex(index, valueCount);
-    UnionListReaderBoundsChecker.checkOffsetBuffer(vector.getOffsetBuffer(), index, OFFSET_WIDTH);
+    UnionListReaderPositionValidator.checkIndex(index, valueCount);
+    UnionListReaderPositionValidator.checkOffsetBufferReadable(
+        vector.getOffsetBuffer(), index, OFFSET_WIDTH);
 
     super.setPosition(index);
     currentOffset = vector.getElementStartIndex(index) - 1;
--- a/vector/src/main/java/org/apache/arrow/vector/complex/impl/UnionLargeListReader.java
+++ b/vector/src/main/java/org/apache/arrow/vector/complex/impl/UnionLargeListReader.java
@@ -60,7 +60,7 @@ public class UnionLargeListReader extends AbstractFieldReader {
     }
 
     UnionListReaderPositionValidator.checkIndex(index, valueCount);
-    UnionListReaderPositionValidator.checkOffsetBufferReadable(
+    UnionListReaderPositionValidator.checkListBufferReadable(
         vector.getOffsetBuffer(), index, OFFSET_WIDTH);
 
     super.setPosition(index);
--- a/vector/src/main/java/org/apache/arrow/vector/complex/impl/UnionLargeListReader.java
+++ b/vector/src/main/java/org/apache/arrow/vector/complex/impl/UnionLargeListReader.java
@@ -54,7 +54,9 @@ public class UnionLargeListReader extends AbstractFieldReader {
   @Override
   public void setPosition(int index) {
     int valueCount = vector.getValueCount();
-    if (valueCount == 0 && index == 0) {
+    if (valueCount == 0
+        && index == 0
+        && vector.getOffsetBuffer().capacity() < 2L * OFFSET_WIDTH) {
       setEmptyPosition(index);
       return;
     }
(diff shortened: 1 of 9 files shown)


bodduv:
I don't think this is a `breaking-change`, it should be labeled `bug-fix`

vector/src/main/java/org/apache/arrow/vector/complex/impl/UnionLargeListReader.java
```diff
@@ -53,9 +53,24 @@ public boolean isSet() {
...
+    super.setPosition(index);
+    currentOffset = vector.getElementStartIndex(index) - 1;
+    maxOffset = vector.getElementEndIndex(index);
+  }
+
+  private void setEmptyPosition(int index) {
     super.setPosition(index);
-    currentOffset = vector.getOffsetBuffer().getLong((long) index * OFFSET_WIDTH) - 1;
-    maxOffset = vector.getOffsetBuffer().getLong(((long) index + 1L) * OFFSET_WIDTH);
+    currentOffset = 0;
```
jhrotko, line 73: `isEmptyVectorPosition` conflates two things: "the vector is empty" (`valueCount == 0`) and "index is `0`." If `valueCount == 0`, why is `index == 0` special at all? Any access on an empty vector should throw, OR the guard should be just `valueCount == 0` (allowing `setPosition(n)` on empty vectors to succeed for any `n`). The current behavior silently succeeds only for index `0` and throws for any other index on an empty vector, which is asymmetric and surprising.

vector/src/main/java/org/apache/arrow/vector/complex/impl/UnionListReader.java
```diff
@@ -52,14 +52,24 @@ public boolean isSet() {
...
-    } else {
-      currentOffset = vector.getOffsetBuffer().getInt(index * (long) OFFSET_WIDTH) - 1;
-      maxOffset = vector.getOffsetBuffer().getInt((index + 1) * (long) OFFSET_WIDTH);
+    int valueCount = vector.getValueCount();
+    if (UnionListReaderBoundsChecker.isEmptyVectorPosition(index, valueCount)) {
+      setEmptyPosition(index);
+      return;
     }
+
+    UnionListReaderBoundsChecker.checkIndex(index, valueCount);
```
jhrotko: you could further move both these methods (`setPosition` and `setEmpty` to he common packcage as well and make a single call here)

vector/src/test/java/org/apache/arrow/vector/ipc/TestRoundTrip.java
```diff
@@ -326,6 +329,82 @@ public void testMetadata(String name, IpcOption writeOption) throws Exception {
...
+  private void validateEmptyUnionListReaders(int expectedCount, VectorSchemaRoot root) {
+    assertEquals(0, expectedCount);
+
+    FieldReader structReader = root.getVector("struct").getReader();
+    assertEmptyUnionListReader(structReader.reader("list"));
+    assertEmptyUnionListReader(structReader.reader("largeList"));
+    assertEmptyUnionListReader(structReader.reader("map"));
+  }
+
+  private void assertEmptyUnionListReader(FieldReader reader) {
```
jhrotko, line 403: `assertEmptyUnionListReader` never calls `reader.setPosition(0)`, which is the exact scenario that was broken.
The test should include `reader.setPosition(0)` followed by `assertEquals(0, reader.size())` and `assertFalse(reader.next())`, otherwise you're not actually verifying the bug is fixed, just that the initial field defaults happen to be 0.

vector/src/main/java/org/apache/arrow/vector/complex/impl/UnionListReaderBoundsChecker.java
```diff
@@ -0,0 +1,47 @@
...
+package org.apache.arrow.vector.complex.impl;
+
+import org.apache.arrow.memory.ArrowBuf;
+
+/** Shared position validation for union list readers backed by offset buffers. */
+final class UnionListReaderBoundsChecker {
+
+  private UnionListReaderBoundsChecker() {}
+
+  static boolean isEmptyVectorPosition(int index, int valueCount) {
```
jhrotko: The method name implies it checks whether a specific position holds an empty list. It actually checks whether the vector itself is empty (`valueCount == 0`). A name like `isEmptyVector` or `shouldBypassEmptyVector` would be clearer

jbonofre (commented):
I guess `UnionListViewReader` needs the same fix, right?

vector/src/main/java/org/apache/arrow/vector/complex/impl/UnionListReaderPositionValidator.java
```diff
@@ -0,0 +1,47 @@
...
+  private UnionListReaderBoundsChecker() {}
+
+  static boolean isEmptyVectorPosition(int index, int valueCount) {
+    return valueCount == 0 && index == 0;
+  }
+
+  static void checkIndex(int index, int valueCount) {
+    if (index < 0 || index >= valueCount) {
+      throw new IndexOutOfBoundsException(
+          String.format("index: %s, expected range (0, %s)", index, valueCount));
```
jbonofre: nit: I read `(0, %s)` as "open", 0 excluded.

I guess it should be `[0, %s)`, or simply `index %d out of bounds for length %d`.

jbonofre (approved):

vector/src/main/java/org/apache/arrow/vector/complex/impl/UnionLargeListReader.java
bodduv: I reworked this.

vector/src/main/java/org/apache/arrow/vector/complex/impl/UnionListReaderBoundsChecker.java
bodduv: I removed this.

vector/src/main/java/org/apache/arrow/vector/complex/impl/UnionListReader.java
bodduv: could you check again?

jhrotko (approved):
LGTM

lidavidm:
@bodduv can you rebase? It seems CI did not trigger

bodduv:
Thank you for taking a look, @lidavidm. I make the PR uptodate.

PR #1137, opened 2026-05-06
GH-1109: Fix ComplexCopier for extension-type holder readers

### LINKED ISSUE #1109: [Vector][ExtensionType] ComplexCopier requires that extension types' holder readers implement getField

https://github.com/apache/arrow-java/pull/892 made efforts to streamline and simplify extension readers and writers. In this, `ComplexCopier.copy` relies on `FieldReader.getField` being defined for extension types.

Usages of `ComplexCopier.copy` with readers backed by vector, e.g., `UuidReaderImpl` and `VariantReaderImpl` that do have `getField` implementation, should be OK (as there is concrete notion of Field). However, if holder readers, namely, `NullableUuidHolderReaderImpl` or `NullableVariantHolderReaderImpl` are used this results in an NPE.

Having said that `getField` method for holder readers appears to be a misnomer as we do not have a notion of a field for the holders. This can be seen from the absence of `getField` overrides for primitive types.

bodduv:
### Background

`ComplexCopier.copy` does a deep-copy using a `FieldReader` from the source and a `FieldWriter` to write to the destination. On the read side in the `EXTENSIONTYPE` branch, it uses `reader.getField().getType()`  to get the `ArrowType` it needs to forward to `writer.writeExtension(value, type)`.  Vector-backed extension readers (UuidReaderImpl, VariantReaderImpl) override getField(), so this works. Holder readers (NullableUuidHolderReaderImpl, NullableVariantHolderReaderImpl) on the other hand do not. `AbstractFieldReader.getField` throws via `fail("getField")`, so any ComplexCopier.copy call with a holder reader as input fails.

As #1109 points out, overriding getField() on a holder reader is a simple fix but a holder does not have a notion of `Field`, the same way primitive (as opposed to extension) holder readers also do not have notion of `getField()` and the method is not overridden.

### What's Changed

`ComplexCopier` only needs the `ArrowType` here, not the whole Field. So this PR adds a narrower accessor for that:

```java
  default ArrowType getExtensionType() {
    return getField().getType();
  }
```

on `ExtensionReader`. The default delegates to `getField()`, so every existing vector-backed extension reader keeps working unchanged. Holder readers override `getExtensionType()` to return `holder.type()` — the `ExtensionHolder.type()` method added in #892 already returns the right ArrowType (UuidType.INSTANCE, VariantType.INSTANCE).

ComplexCopier call `reader.getExtensionType()` instead of `reader.getField().getType()`. The new method goes on ExtensionReader rather than AbstractFieldReader to sit next to the other extension-specific methods already there (read(ExtensionHolder), readObject(), isSet()). Adding a default method is this specific way is source- and binary-compatible.

Closes #1109 .

### DIFF
--- a/arrow-variant/src/main/java/org/apache/arrow/variant/impl/NullableVariantHolderReaderImpl.java
+++ b/arrow-variant/src/main/java/org/apache/arrow/variant/impl/NullableVariantHolderReaderImpl.java
@@ -19,6 +19,7 @@ package org.apache.arrow.variant.impl;
 import org.apache.arrow.variant.holders.NullableVariantHolder;
 import org.apache.arrow.vector.complex.impl.AbstractFieldReader;
 import org.apache.arrow.vector.types.Types;
+import org.apache.arrow.vector.types.pojo.ArrowType;
 
 public class NullableVariantHolderReaderImpl extends AbstractFieldReader {
   private final NullableVariantHolder holder;
@@ -45,6 +46,11 @@ public class NullableVariantHolderReaderImpl extends AbstractFieldReader {
   @Override
   public Types.MinorType getMinorType() {
     return Types.MinorType.EXTENSIONTYPE;
+  }
+
+  @Override
+  public ArrowType getExtensionType() {
+    return holder.type();
   }
 
   @Override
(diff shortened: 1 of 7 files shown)


jbonofre (approved):
LGTM, Thanks!

bodduv:
Pointing Github "Explain Error" on failing build yielded:

> Looking at the recent RC workflow runs on main:
> - Run 24976298831 (main branch, RC workflow): FAILED - Same arrow::decimal compilation error

> This shows the issue is pre-existing in main, not introduced by PR #1137. The JNI compilation failure affects:
> - The RC (Release Candidate) workflow on macos-14 aarch64
> - Both on main branch and PR #1137

Some runs passed for PR #1137 though.

jhrotko (approved):
Thanks for taking care of this!

bodduv:
@jbonofre there might be an issue with CI, could we try to rebuild?

jbonofre:
I'm working on CI fix right now.

lidavidm:
@bodduv  please rebase

bodduv:
Thank you for taking a look, @lidavidm. I make the PR up-to-date.

PR #1139, opened 2026-05-06
GH-1134: add COLUMN_DEF support to JDBC getColumnns()

### LINKED ISSUE #1134: Consume COLUMN_DEF in JDBC Flight Driver

`ArrowDatabaseMetadata`'s schema currently defines a vector for `COLUMN_DEF`, yet never writes any value to it.  `FlightSQLColumnMetadata` also does not have a constant/getter/setter for this value.

Due to this, database connections through java's arrow-flight cannot and will not show underlying default value metadata per column.  DESCRIBE TABLE or similar calls (engine depending) can still be used to get a column's default value, but it can be at odds with what the client shows.

I confirmed this behavior using Dremio as the query engine, dBeaver as the client, and iceberg as the table containing a column with a default value.

DBeaver and other DB clients should accurately show column default values when calling DatabaseMetaData.getColumns() public method.  Clients will need to send this information, and the arrow driver needs to consume this information.

Verest:
## What's Changed

Updates the `FlightSqlColumnMetadata` to contain the JDBC `COLUMN_DEF` field alongside getters/setters.  Following exiting naming format.

Updates `ArrowDatabaseMetadata` to write to the pre-existing `COLUMN_DEF` vector as part of the `getColumns()` code paths.

Updates test(s) in `ArrowDatabaseMetadataTest` to test default value behavior.

Closes #1134.

### DIFF
--- a/arrow-format/FlightSql.proto
+++ b/arrow-format/FlightSql.proto
@@ -1213,6 +1213,7 @@ message CommandGetDbSchemas {
  *  - ARROW:FLIGHT:SQL:IS_READ_ONLY      - "1" indicates if the column is read only, "0" otherwise.
  *  - ARROW:FLIGHT:SQL:IS_SEARCHABLE     - "1" indicates if the column is searchable via WHERE clause, "0" otherwise.
  *  - ARROW:FLIGHT:SQL:REMARKS           - A comment describing the column.
+ *  - ARROW:FLIGHT:SQL:COLUMN_DEF        - The default value for the column.
  * The returned data should be ordered by catalog_name, db_schema_name, table_name, then table_type, followed by table_schema if requested.
  */
 message CommandGetTables {
--- a/arrow-format/FlightSql.proto
+++ b/arrow-format/FlightSql.proto
@@ -1681,6 +1681,7 @@ message ActionEndSavepointRequest {
  *    - ARROW:FLIGHT:SQL:IS_READ_ONLY      - "1" indicates if the column is read only, "0" otherwise.
  *    - ARROW:FLIGHT:SQL:IS_SEARCHABLE     - "1" indicates if the column is searchable via WHERE clause, "0" otherwise.
  *    - ARROW:FLIGHT:SQL:REMARKS           - A comment describing the column.
+ *    - ARROW:FLIGHT:SQL:COLUMN_DEF        - The default value for the column.
  *  - GetFlightInfo: execute the query.
  */
 message CommandStatementQuery {
@@ -1707,6 +1708,7 @@ message CommandStatementQuery {
  *    - ARROW:FLIGHT:SQL:IS_READ_ONLY      - "1" indicates if the column is read only, "0" otherwise.
  *    - ARROW:FLIGHT:SQL:IS_SEARCHABLE     - "1" indicates if the column is searchable via WHERE clause, "0" otherwise.
  *    - ARROW:FLIGHT:SQL:REMARKS           - A comment describing the column.
+ *    - ARROW:FLIGHT:SQL:COLUMN_DEF        - The default value for the column.
  *  - GetFlightInfo: execute the query.
  *  - DoPut: execute the query.
  */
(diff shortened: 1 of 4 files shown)


Verest:
I am murky on the specific format for `COLUMN_DEF`, or where that responsibility lies.  The Oracle [docs](https://docs.oracle.com/javase/8/docs/api/java/sql/DatabaseMetaData.html) mention wrapping the underling value in single quotes if it is a string, and a few other things.  I suspect this format is not followed everywhere though 🤔.

Since this is a generic JDBC driver, I am cautious to make assumptions about the incoming format of the `COLUMN_DEF` from the server.  For now, I am tentatively expecting the server-side to handle the intended parsing of the `COLUMN_DEF`, leaving the arrow driver to read given strings as is.

Open to suggestions here.

Aside: this PR should be an enhancement, but I do not have permission to edit the labels.

xborder:
Should this be discussed and voted like it was done for [is_update](https://lists.apache.org/thread/88msflnwkkw8t81czs2ndqhkn1fb1pxd)? Maybe it is overkill but I think at least [apache/arrow/FlightSql.proto](https://github.com/apache/arrow/blob/main/format/FlightSql.proto) should be in sync.

There are other places in FlightSql.proto that also mention metadata returned that probably would need to be updated

xborder:
Have you thought if it would make sense in [CommandGetXdbcTypeInfo](https://github.com/apache/arrow/blob/2a89d03bbefd620b42126b8e00f8ae57e99cd638/format/FlightSql.proto#L1132)?

ennuite:
> Should this be discussed and voted like it was done for [is_update](https://lists.apache.org/thread/88msflnwkkw8t81czs2ndqhkn1fb1pxd)? Maybe it is overkill but I think at least [apache/arrow/FlightSql.proto](https://github.com/apache/arrow/blob/main/format/FlightSql.proto) should be in sync.
> 
> There are other places in FlightSql.proto that also mention metadata returned that probably would need to be updated

Yes, the proto files in this repo should mirror the main one.

arrow-format/FlightSql.proto
```diff
@@ -1213,6 +1213,7 @@ message CommandGetDbSchemas {
  *  - ARROW:FLIGHT:SQL:IS_READ_ONLY      - "1" indicates if the column is read only, "0" otherwise.
  *  - ARROW:FLIGHT:SQL:IS_SEARCHABLE     - "1" indicates if the column is searchable via WHERE clause, "0" otherwise.
  *  - ARROW:FLIGHT:SQL:REMARKS           - A comment describing the column.
+ *  - ARROW:FLIGHT:SQL:COLUMN_DEF        - The default value for the column.
```
jbonofre, line 1216: Thanks to have update the javadoc here.

I think it's worth to document `COLUMN_DEF` in the javadoc of `CommandStatementQuery`, `CommandStatementSubstraitPlan`, `CommandPreparedStatementQuery`. As we have `REMARKS` in these messages, we should have `COLUMN_DEF`.

jbonofre:
By the way, as this PR touches `.proto`, worth to at discuss on the mailing list.

arrow-format/FlightSql.proto
Verest: Yes, I was uncertain about that - I can add them.  

I did not add them originally as I do not think the COLUMN_DEF will be used in those code paths, but I don't think it hurts having it there either.

Verest:
> Have you thought if it would make sense in [CommandGetXdbcTypeInfo](https://github.com/apache/arrow/blob/2a89d03bbefd620b42126b8e00f8ae57e99cd638/format/FlightSql.proto#L1132)?

Off-hand I would say a defaultValue is N/A for that class as it appears to be generic to SQL types rather than specific columns - A default value is only relevant for a specific column of a table.

Maybe something along the lines of "supports_default_values" could be applicable there, but that would be out of scope for this PR.  I am murky on the specifics of that class regardless.

lidavidm (approved):
It seems reasonable to me. We should proceed with a vote on ML

PR #1140, opened 2026-05-07
MINOR: Fix Gandiva JNI build against Arrow C++ 24.0.0

jbonofre:
Arrow C++ 24.0.0 introduced two breaking changes for the JNI build:

1. `arrow::decimal()` was removed; replaced with `arrow::decimal128()` in the Gandiva JNI source.

2. xsimd >= 14.0.0 is now a required dependency. The Docker image's vcpkg registry only ships xsimd 13.2.0, so the Linux JNI CMake configuration was failing. Fixed by passing `xsimd_SOURCE=BUNDLED` to the Arrow C++ CMake configure step so Arrow downloads and uses xsimd 14.0.0 directly, and by passing the vcpkg toolchain file to the Arrow C++ configure step so other vcpkg-managed dependencies are still resolved correctly.

### DIFF
--- a/ci/scripts/jni_manylinux_build.sh
+++ b/ci/scripts/jni_manylinux_build.sh
@@ -74,10 +74,7 @@ cmake \
   -S "${arrow_dir}/cpp" \
   -B "${build_dir}/cpp" \
   --preset=ninja-release-jni-linux \
-  -DCMAKE_INSTALL_PREFIX="${install_dir}" \
-  -DCMAKE_TOOLCHAIN_FILE="${VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake" \
-  -DVCPKG_TARGET_TRIPLET="${VCPKG_TARGET_TRIPLET}" \
-  -Dxsimd_SOURCE=BUNDLED
+  -DCMAKE_INSTALL_PREFIX="${install_dir}"
 cmake --build "${build_dir}/cpp"
 cmake --install "${build_dir}/cpp"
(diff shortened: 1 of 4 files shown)


lidavidm (approved):

ci/scripts/jni_manylinux_build.sh
```diff
@@ -74,7 +74,10 @@ cmake \
   -S "${arrow_dir}/cpp" \
   -B "${build_dir}/cpp" \
   --preset=ninja-release-jni-linux \
-  -DCMAKE_INSTALL_PREFIX="${install_dir}"
+  -DCMAKE_INSTALL_PREFIX="${install_dir}" \
+  -DCMAKE_TOOLCHAIN_FILE="${VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake" \
+  -DVCPKG_TARGET_TRIPLET="${VCPKG_TARGET_TRIPLET}" \
+  -Dxsimd_SOURCE=BUNDLED
```
kou: Could you update https://github.com/apache/arrow-java/blob/0f0a58433b46004ff9869a2512beded177340bca/.env#L56 to synchronize with https://github.com/apache/arrow/blob/a0d2885b101acb439f7f79ec2237028974e74e64/.env#L95 instead of `-Dxsimd_SOURCE=BUNDLED`?

jbonofre: Good point. Thanks!

PR #1141, opened 2026-05-07
GH-1067: Close cached HDFS FileSystem instances

### LINKED ISSUE #1067: [ARROW Java][HDFS] JVM hangs after reading HDFS files via Arrow Dataset API due to non-daemon native threads
When using the Apache Arrow Java Dataset API (FileSystemDatasetFactory) to read ORC files directly from HDFS, the JVM fails to exit gracefully after the reading process is complete. The application hangs indefinitely because two non-daemon native threads (started by the underlying libhdfs JNI layer) remain active.

This issue does not occur when reading local files; the JVM exits normally in that scenario. It specifically affects HDFS interactions where the C++ libhdfs client is loaded via JNI.

I have tested this on Apache Arrow version 9.0.0 (the last version with built-in HDFS support) and confirmed the behavior persists. I also attempted version 17.0.0, but the thread leakage behavior remains the same once the dependency is added.

Currently, the only workaround is to force terminate the JVM using System.exit(0), which is not ideal for applications relying on shutdown hooks or running within complex containers.

here is java code:  
String hdfsUri = "hdfs://node01.cdh5:8020/user/hive/warehouse/perf_test_200col_500w_nopk/ds=2/perf_test_200col_500w_nopk";
        ScanOptions options = new ScanOptions(/*batchSize*/ 32768);
        try (
            BufferAllocator allocator = new RootAllocator();
            NativeMemoryPool pool = NativeMemoryPool.getDefault();
            DatasetFactory datasetFactory = new FileSystemDatasetFactory(allocator, pool, FileFormat.ORC, hdfsUri);
            Dataset dataset = datasetFactory.finish();
            Scanner scanner = dataset.newScan(options);
            ArrowReader reader = scanner.scanBatches()
        ) {
            int totalBatchSize = 0;
            while (reader.loadNextBatch()) {
                try (VectorSchemaRoot root = reader.getVectorSchemaRoot()) {
                    totalBatchSize += root.getRowCount();
                }
            }
            System.out.println("Total batch size: " + totalBatchSize);
        } catch (Exception e) {
            e.printStackTrace();
        }

xborder:
## What's Changed

- This PR fixes JVM shutdown hangs after reading HDFS datasets through Arrow Java.
- FileSystemDatasetFactory now tracks hdfs:// URIs used to create the factory. On close(), after releasing the native dataset factory, it best-effort closes the matching Hadoop FileSystem instances.
- The Hadoop cleanup is done via reflection so Arrow Java does not add a production dependency on Hadoop. Non-HDFS URIs are ignored.

Closes #1067 .

### DIFF
--- a/dataset/src/main/java/org/apache/arrow/dataset/file/FileSystemDatasetFactory.java
+++ b/dataset/src/main/java/org/apache/arrow/dataset/file/FileSystemDatasetFactory.java
@@ -16,7 +16,11 @@
  */
 package org.apache.arrow.dataset.file;
 
+import java.lang.reflect.Method;
+import java.net.URI;
+import java.util.LinkedHashSet;
 import java.util.Optional;
+import java.util.Set;
 import org.apache.arrow.dataset.jni.NativeDatasetFactory;
 import org.apache.arrow.dataset.jni.NativeMemoryPool;
 import org.apache.arrow.dataset.scanner.FragmentScanOptions;
@@ -25,9 +29,12 @@ import org.apache.arrow.memory.BufferAllocator;
 /** Java binding of the C++ FileSystemDatasetFactory. */
 public class FileSystemDatasetFactory extends NativeDatasetFactory {
 
+  private final String[] uris;
+
   public FileSystemDatasetFactory(
       BufferAllocator allocator, NativeMemoryPool memoryPool, FileFormat format, String uri) {
     super(allocator, memoryPool, createNative(format, uri, Optional.empty()));
+    this.uris = uri == null ? new String[0] : new String[] {uri};
   }
 
   public FileSystemDatasetFactory(
@@ -37,11 +44,13 @@ public class FileSystemDatasetFactory extends NativeDatasetFactory {
       String uri,
       Optional<FragmentScanOptions> fragmentScanOptions) {
     super(allocator, memoryPool, createNative(format, uri, fragmentScanOptions));
+    this.uris = uri == null ? new String[0] : new String[] {uri};
   }
 
   public FileSystemDatasetFactory(
       BufferAllocator allocator, NativeMemoryPool memoryPool, FileFormat format, String[] uris) {
     super(allocator, memoryPool, createNative(format, uris, Optional.empty()));
+    this.uris = uris == null ? new String[0] : uris.clone();
   }
 
   public FileSystemDatasetFactory(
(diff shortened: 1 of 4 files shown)


jbonofre:
@xborder I fixed the CI on main. Can you please rebase?

lidavidm:
Sorry, probably need to rebase again.

xborder:
done @lidavidm

dataset/src/main/java/org/apache/arrow/dataset/file/FileSystemDatasetFactory.java
```diff
@@ -51,6 +60,65 @@ public FileSystemDatasetFactory(
...
+   * Hadoop FileSystem to release non-daemon threads that would otherwise prevent JVM exit. See <a
+   * href="https://github.com/apache/arrow-java/issues/1067">#1067</a>.
+   */
+  @Override
+  public synchronized void close() {
+    try {
+      super.close();
+    } finally {
+      hdfsFileSystems.forEach(FileSystemDatasetFactory::closeHadoopFileSystem);
+    }
```
lidavidm, line 82: nit: could we use AutoCloseables here?

dataset/src/main/java/org/apache/arrow/dataset/file/FileSystemDatasetFactory.java
```diff
@@ -51,6 +60,65 @@ public FileSystemDatasetFactory(
...
+    try {
+      Class<?> confClass = Class.forName("org.apache.hadoop.conf.Configuration");
+      Object conf = confClass.getDeclaredConstructor().newInstance();
+      Class<?> fsClass = Class.forName("org.apache.hadoop.fs.FileSystem");
+      Method getMethod = fsClass.getMethod("get", URI.class, confClass);
+      Object fs = getMethod.invoke(null, hdfsUri, conf);
+      Method closeMethod = fsClass.getMethod("close");
+      closeMethod.invoke(fs);
+    } catch (Exception e) {
+      // Best-effort cleanup; Hadoop may not be on classpath or FileSystem already closed
```
lidavidm: nit: worth perhaps logging this?

dataset/src/main/java/org/apache/arrow/dataset/file/FileSystemDatasetFactory.java
```diff
@@ -51,6 +60,65 @@ public FileSystemDatasetFactory(
...
+    }
+    for (String uri : uris) {
+      try {
+        URI parsedUri = new URI(uri);
+        if ("hdfs".equalsIgnoreCase(parsedUri.getScheme())) {
+          hdfsFileSystems.add(
+              new URI(parsedUri.getScheme(), parsedUri.getAuthority(), null, null, null));
+        }
+      } catch (Exception e) {
+        // Ignore here; native factory creation reports invalid user URIs.
```
lidavidm: nit: could we catch a more specific exception class for invalid URIs?

dataset/src/main/java/org/apache/arrow/dataset/file/FileSystemDatasetFactory.java
xborder: done. added a debug log

dataset/src/main/java/org/apache/arrow/dataset/file/FileSystemDatasetFactory.java
xborder: done. now avoids null uris and catches `URISyntaxException` specifically

dataset/src/main/java/org/apache/arrow/dataset/file/FileSystemDatasetFactory.java
xborder: I think we could use AutoCloseables by casting the `Object` obtained through reflection, but it would not simplify or improve the cleanup logic.

Is there any benefit in changing it that I'm not seeing?

lidavidm: Well, you would assemble a list of lambdas essentially and close it all at once; that way a failure to close one won't affect the others. But this is fine.

lidavidm (approved):

PR #1145, opened 2026-05-11
GH-1142: Fix buffer ownership in VectorSchemaRoot addVector/removeVector

### LINKED ISSUE #1142: Managing ownership in VectorSchemaRoot#addVector, recent changes miss the main fault.
A recent change in VectorSchemaRoot#addVector provided a simple improvement, but completely misses a major shortcoming of this routine and, almost certainly, others like it. Based on the interest in that issue (#301), this issue might be of interest to @axreldable and @jbonofre.

One cannot change the `FieldVector`s belonging to `VectorSchemaRoot`s by simply rearranging the elements of their `List`s without considering the ownership of their underlying buffers. This can cause various problems, particularly when `VectorSchemaRoot`s are closed. This is why `TransferPair` exists.

I would probably need some help with improving this as I'm quite new to using Apache Arrow (and very new to using GitHub). But I have written a possibly improved implementation if this would be of interest and would be open to collaborating on rectifying this bug.

John-W-Lewis:
## What's Changed

`VectorSchemaRoot#addVector` and `#removeVector` previously created a new `VectorSchemaRoot` by sharing raw `FieldVector` references between the original and new root. This violates Arrow's ownership model: closing either root would release buffers still in use by the other, leading to use-after-free errors.

This fix uses `TransferPair` to properly transfer buffer ownership to the returned root, consistent with how `VectorSchemaRoot#slice` already handles this. After the operation, the original root's vectors are left in a transferred (empty) state and can be reused via `allocateNew()`.

Tests added to verify that the returned root's data remains valid after the original root and input vector are closed.

Closes #1142.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/vector/src/main/java/org/apache/arrow/vector/VectorSchemaRoot.java
+++ b/vector/src/main/java/org/apache/arrow/vector/VectorSchemaRoot.java
@@ -193,6 +193,10 @@ public class VectorSchemaRoot implements AutoCloseable {
   /**
    * Add vector to the record batch, producing a new VectorSchemaRoot.
    *
+   * <p>Buffer ownership is transferred to the returned root via {@link TransferPair}. After this
+   * operation, the vectors in this root and the added vector are left in a transferred (empty)
+   * state. This root can be reused by calling {@link #allocateNew()}.
+   *
    * @param index field index
    * @param vector vector to be added.
    * @return out VectorSchemaRoot with vector added
@@ -201,22 +205,32 @@ public class VectorSchemaRoot implements AutoCloseable {
     Preconditions.checkNotNull(vector);
     Preconditions.checkArgument(index >= 0 && index <= fieldVectors.size());
     List<FieldVector> newVectors = new ArrayList<>();
-    if (index == fieldVectors.size()) {
-      newVectors.addAll(fieldVectors);
-      newVectors.add(vector);
-    } else {
-      for (int i = 0; i < fieldVectors.size(); i++) {
-        if (i == index) {
-          newVectors.add(vector);
-        }
-        newVectors.add(fieldVectors.get(i));
+    for (int i = 0; i < fieldVectors.size(); i++) {
+      if (i == index) {
+        TransferPair addPair = vector.getTransferPair(vector.getAllocator());
+        addPair.transfer();
+        newVectors.add((FieldVector) addPair.getTo());
       }
+      FieldVector v = fieldVectors.get(i);
+      TransferPair transferPair = v.getTransferPair(v.getAllocator());
+      transferPair.transfer();
+      newVectors.add((FieldVector) transferPair.getTo());
+    }
+    if (index == fieldVectors.size()) {
+      TransferPair addPair = vector.getTransferPair(vector.getAllocator());
+      addPair.transfer();
+      newVectors.add((FieldVector) addPair.getTo());
     }
     return new VectorSchemaRoot(newVectors);
   }
 
   /**
    * Remove vector from the record batch, producing a new VectorSchemaRoot.
+   *
+   * <p>Buffer ownership is transferred to the returned root via {@link TransferPair}. After this
+   * operation, the vectors in this root are left in a transferred (empty) state. The removed
+   * vector's data is not transferred and is released. This root can be reused by calling {@link
+   * #allocateNew()}.
    *
    * @param index field index
    * @return out VectorSchemaRoot with vector removed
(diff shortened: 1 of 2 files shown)


John-W-Lewis:
Being neither a chicken nor an egg, I understand that I am unable to label this as "bug-fix".
Thank you to whoever can do that.

John-W-Lewis:
There are also a couple of enhancements I would suggest for VectorSchemaRoot but, based on the advice I've seen, it seems better to describe those in a separate PR.

axreldable:
Hi @John-W-Lewis , thank you for the bug-fix. Yes, we need to wait for the maintainers to add the label and trigger the other checks. In the meantime, you can build locally, for example using [Maven](https://arrow.apache.org/docs/developers/java/building.html#id2). Now, there is a format violation in `TestVectorSchemaRoot.java`.

```
mvn clean install
...
[ERROR] Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:3.4.0:check (spotless-check) on project arrow-vector: The following files had format violations:
[ERROR]     src/test/java/org/apache/arrow/vector/TestVectorSchemaRoot.java
[ERROR]         @@ -363,8 +363,7 @@
[ERROR]          ······intVector2.setValueCount(5);
[ERROR]          ······intVector3.setValueCount(5);
[ERROR]          
[ERROR]         -······VectorSchemaRoot·original·=
[ERROR]         -··········new·VectorSchemaRoot(Arrays.asList(intVector1,·intVector2));
[ERROR]         +······VectorSchemaRoot·original·=·new·VectorSchemaRoot(Arrays.asList(intVector1,·intVector2));
[ERROR]          ······original.setRowCount(5);
[ERROR]          
[ERROR]          ······VectorSchemaRoot·result·=·original.addVector(1,·intVector3);
[ERROR] Run 'mvn spotless:apply' to fix these violations.
```

You can fix it with the suggested command. `mvn spotless:apply`

image

With the formatting fix, the build was successful for me, which means that the unit tests passed.

vector/src/main/java/org/apache/arrow/vector/VectorSchemaRoot.java
```diff
@@ -226,7 +240,10 @@ public VectorSchemaRoot removeVector(int index) {
     List<FieldVector> newVectors = new ArrayList<>();
     for (int i = 0; i < fieldVectors.size(); i++) {
       if (i != index) {
-        newVectors.add(fieldVectors.get(i));
+        FieldVector v = fieldVectors.get(i);
+        TransferPair transferPair = v.getTransferPair(v.getAllocator());
+        transferPair.transfer();
+        newVectors.add((FieldVector) transferPair.getTo());
```
axreldable: nit: Consider extracting the transferring logic into a separate method to avoid code duplication.
e.g.
```
private static FieldVector transferVector(FieldVector vector) {
    TransferPair transferPair = vector.getTransferPair(vector.getAllocator());
    transferPair.transfer();
    return (FieldVector) transferPair.getTo();
  }
```

vector/src/main/java/org/apache/arrow/vector/VectorSchemaRoot.java
```diff
@@ -201,23 +205,33 @@ public VectorSchemaRoot addVector(int index, FieldVector vector) {
...
     }
     return new VectorSchemaRoot(newVectors);
   }
 
   /**
    * Remove vector from the record batch, producing a new VectorSchemaRoot.
    *
+   * <p>Buffer ownership is transferred to the returned root via {@link TransferPair}. After this
+   * operation, the vectors in this root are left in a transferred (empty) state. The removed
+   * vector's data is not transferred and is released. This root can be reused by calling {@link
```
axreldable, line 225: > The removed vector's data is not transferred and is released.

`is released` is not 100% accurate and can be confusing. The vector remains in the original schema root until it's closed/cleared.

axreldable (approved):
Other than the nit (nitpick) and the doc clarification, LGTM.

John-W-Lewis:
@axreldable Thank you. TestVectorSchemaRoot.java formatting fixed.

vector/src/main/java/org/apache/arrow/vector/VectorSchemaRoot.java
John-W-Lewis: Good suggestion., done.

vector/src/main/java/org/apache/arrow/vector/VectorSchemaRoot.java
John-W-Lewis: Thank you for spotting this. You're right that "is released" was inaccurate.

Looking more closely, it was also a code issue. For the source vectors which are not removed, transfer() creates new ArrowBufs in the corresponding target vector, each referencing the same underlying allocation as its source counterpart, while the source vector's ArrowBufs release their references. So, after the operation, those source vectors' ArrowBufs no longer own any allocated memory. But for the removed vector, we weren't doing anything -- its ArrowBufs still held references to their underlying allocations, which would only be freed when the original root was eventually closed.

I've added a clear() on the removed vector so that its buffers also release their underlying allocations, making it consistent with the other vectors in the source root after the operation.

John-W-Lewis:
Closing in favour of #1149, which provides a better fix using a new `VectorOps.shareCopy` utility. That approach preserves the original intended semantics (both source and result roots remain readable) while making it safe through proper reference counting, rather than changing the semantics to transfer/empty the source.

Thank you @axreldable for your helpful review feedback on this PR -- it contributed to the thinking that led to the improved solution.

Issue #834, opened 2025-08-31
Avro adapter: AvroNullableConsumer should use readIndex


sap1ens:
### Describe the bug, including details regarding any error messages, version, and platform.

When using Avro's `ValidatingDecoder`, any schema with a nullable field fails with the following exception:

```
Exception in thread "main" org.apache.avro.AvroTypeException: Attempt to process a int when a union was expected.
	at org.apache.avro.io.parsing.Parser.advance(Parser.java:91)
	at org.apache.avro.io.ValidatingDecoder.readInt(ValidatingDecoder.java:81)
	at org.apache.arrow.adapter.avro.consumers.AvroNullableConsumer.consume(AvroNullableConsumer.java:44)
```

Reason: `AvroNullableConsumer` uses `readInt` to read the index, but it should use `readIndex` instead. It's irrelevant in case of a `BinaryDecoder`, for example (`readIndex` is the same as `readInt`), but `ValidatingDecoder`s and any Decoders based on that have different logic.

PR #1147, opened 2026-05-13
GH-834: Fix Avro nullable consumer to use readIndex for unions

Kaustav-Sarkar:
## Description: 
Updates Avro consumers to use readIndex() for unions and readString() for strings instead of readInt() and readBytes(). This ensures compatibility with Avro's ValidatingDecoder, which strictly enforces type-specific read methods. Includes a regression test for nullable strings.


Kaustav-Sarkar:
Can a maintainer add labels, I do not have permission

PR #1159, opened 2026-05-19
GH-1158: Fix inclusive bounds checking for DictionaryDecoder

### LINKED ISSUE #1158: [Java] DictionaryEncoder doesn't crash when decoding index outside of Dictionary
### Background
When manually creating an encoded vector, with values referencing indices in a Dictionary, it is possible to have values equal to `valueCount` of the Dictionary vector i.e. one index out of bounds. This is interpreted as a null value in the Dictionary.

---

### Example
_Dictionary_
| Index | Original Value |
|-------|---------------|
| 0         | Foo                 |

_Encoded Vector - 1_
| Index | Encoded Value | Expected Decode Outcome             | Actual Outcome |
|-------|----------------|-----------------------------------|-----------------|
| 0         | 0                        | Return the original value                 | Returns the original value |

_Encoded Vector - 2_
| Index | Encoded Value | Expected Decode Outcome             | Actual Outcome |
|-------|----------------|-----------------------------------|-----------------|
| 0         | 1                         | Raise `IllegalArgumentException` | Returns null |

_Encoded Vector - 3_
| Index | Encoded Value | Expected Decode Outcome            | Actual Outcome |
|-------|----------------|-----------------------------------|-----------------|
| 0         | 2                        | Raise `IllegalArgumentException` | Raises `IllegalArgumentException`|

---

### Test to reproduce the error
```java
    @Test
    public void testReferencingIndexOutOfBounds() {
        // Index at which the original value will be stored at in the dictionary
        var encodedIndex = 0;
        // The encoded value that references an index in the dictionary
        var indexReferenced = 1;

        try (final IntVector encodedVector = new IntVector("encodings", allocator);
             final VarCharVector dictionaryVector = newVarCharVector("dict", allocator); ) {
            var originalValue = "Foo";
            dictionaryVector.allocateNew(1);
            dictionaryVector.setValueCount(1);
            dictionaryVector.set(encodedIndex, originalValue.getBytes(StandardCharsets.UTF_8));

            encodedVector.allocateNew(1);
            encodedVector.setValueCount(1);
            encodedVector.set(0, indexReferenced);

            Dictionary dictionary =
                    new Dictionary(dictionaryVector, new DictionaryEncoding(1L, false, null));

            try (ValueVector decoded = DictionaryEncoder.decode(encodedVector, dictionary)) {
                fail("There should be an exception when decoding index outside dictionary's range.");
            } catch (Exception e) {
                assertEquals("Provided dictionary does not contain value for index " + indexReferenced, e.getMessage());
            }
        }
    }

```

jakobhok-neo4j:
Closes #1158.

### DIFF
--- a/vector/src/test/java/org/apache/arrow/vector/TestDictionaryVector.java
+++ b/vector/src/test/java/org/apache/arrow/vector/TestDictionaryVector.java
@@ -403,7 +403,7 @@ public class TestDictionaryVector {
   }
 
   @Test
-  public void testVarcharEquals() {
+  public void testVarCharEquals() {
     try (final VarCharVector vector1 = new VarCharVector("varchar", allocator);
         final VarCharVector vector2 = new VarCharVector("varchar", allocator)) {
 
@@ -1061,7 +1061,7 @@ public class TestDictionaryVector {
       try (final StructVector decode = StructSubfieldEncoder.decode(indices, provider, allocator)) {
         fail("There should be an exception when decoding");
       } catch (Exception e) {
-        assertEquals("Provided dictionary does not contain value for index 3", e.getMessage());
+        assertEquals("Provided dictionary does not contain value for index 1", e.getMessage());
       }
     }
     assertEquals(0, allocator.getAllocatedMemory(), "struct decode memory leak");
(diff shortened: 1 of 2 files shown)


vector/src/test/java/org/apache/arrow/vector/TestDictionaryVector.java
lidavidm, line 1: Could use assertThrows for new tests?

lidavidm (commented):
Seems reasonable. Can you rebase to pick up CI fixes?

PR #1180, opened 2026-06-09
GH-1179: Correct the size of var-width vector with >0 start offset during vector append

### LINKED ISSUE #1179: [Java] Incorrect size computed for var-width vectors with non-zero start offsets when appending them

When appending a variable width vector to another in Java, it computes the size of the currently appended vector by just taking its end offset. It never subtracts the start offset because the majority of start offsets are 0. However, if another language slices a buffer and passes it to Java, it may have a non zero starting offset. This causes us to drastically overestimate the amount of memory to allocate and can cause errors downstream.

jordepic:
## What's Changed

Fix VectorAppender data size computation for variable-width vectors with non-zero start offsets

When appending a variable width offset vector in DataFusion comet I was receiving exceptions
due to allocating too much memory. This is because Comet passes variable width arrays back
to Java where the initial offset vector entry is greater than 0. Prior to this change, arrow-java
determines how many bytes to copy by just looking at the last offset entry in the buffer,
completely disregarding the value of the first.  If first = 100 and last = 200, Java will still
copy 200 bytes instead of 100. In this change we fix that.

Closes #1179

### DIFF
--- a/vector/src/main/java/org/apache/arrow/vector/util/VectorAppender.java
+++ b/vector/src/main/java/org/apache/arrow/vector/util/VectorAppender.java
@@ -340,16 +340,20 @@ public class VectorAppender implements VectorVisitor<ValueVector, Void> {
         targetVector
             .getOffsetBuffer()
             .getInt((long) targetVector.getValueCount() * ListVector.OFFSET_WIDTH);
-    int deltaListSize =
+    // see the corresponding comment in visit(BaseVariableWidthVector, Void): the delta's
+    // offset buffer need not start at zero
+    int deltaListStart = deltaVector.getOffsetBuffer().getInt(0);
+    int deltaListEnd =
         deltaVector
             .getOffsetBuffer()
             .getInt((long) deltaVector.getValueCount() * ListVector.OFFSET_WIDTH);
+    int deltaListSize = deltaListEnd - deltaListStart;
 
     ListVector targetListVector = (ListVector) targetVector;
 
     // make sure the underlying vector has value count set
     targetListVector.getDataVector().setValueCount(targetListSize);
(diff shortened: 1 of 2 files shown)


jordepic:
Could a maintainer please add the bug-fix label here?

jordepic:
Hey @lidavidm , @laurentgo , @wgtmac , @jbonofre - would one of you guys mind taking a look here? It's a pretty nefarious bug :)

jbonofre (approved):
LGTM.

I believe the same issue is present in `ListVector` and `LargeListVector`. Maybe worth to address in this PR or a following up PR.

jordepic:
@jbonofre would you be able to label the PR so that those checks aren't failing? I don't have permissions. I can also follow up here with the list vector changes. Thank you for your review :)

jbonofre:
@jordepic absolutely

jordepic:
I'm not sure that the label worked if you tried to add it, I appreciate all of your help here

jordepic:
@jbonofre I went ahead and amended this guy with the ListVector and LargeListVector fix. Changes are tested.  Would greatly appreciate you labeling the PR as a bug fix as well as a review, thank you :)

jordepic:
Thanks @lidavidm !! I think the last failure here is transient if you want to give it a reroll

jordepic:
Go big red, by the way ;)

jordepic:
@jbonofre if you want to do a final sweep all checks are passing!

jordepic:
@lidavidm would you mind taking a look at this change? I'm hoping to get it through since it can be a slightly dangerous bug!

lidavidm (approved):

PR #1189, opened 2026-06-15
GH-1188: Reduce test workflow waste

### LINKED ISSUE #1188: [CI] Reduce CI test usage

The current Test GitHub Actions workflow spends a large amount of runner time repeatedly compiling the same Maven reactor across the matrix.

- Today, most jobs run `build.sh` followed by `test.sh`. `build.sh` runs a full `mvn clean install -DskipTests`, then `test.sh` runs `mvn clean test`, which deletes the build outputs and recompiles before running tests. 
- JNI/C Data jobs add additional Maven test invocations and native build work, so some jobs compile or build equivalent artifacts multiple times.
- For macOS and Windows runners, Java bytecode is compiled again even though the project targets maven.compiler.release=17, making the compiled classes portable across JDK 17+ and across OS/architecture for pure Java tests.
- Other smaller inefficiencies:
  - The workflow can run twice for PR branch updates because both `push` and `pull_request` triggers fire.f
  - Integration is not required for every PR, but currently participates broadly in the workflow.

This was triggered by https://lists.apache.org/thread/drf0o5kzg1zfmok7gc09k8qz8hh9ymvh

xborder:
## Summary

This PR reduces wasted GitHub Actions time in the `Test` workflow by removing redundant rebuilds, improving caching, and skipping integration work when it is not relevant.

## What changed

- Removed `clean` from `ci/scripts/test.sh` so the test phase reuses the classes compiled earlier in the job instead of deleting and recompiling them.
- Narrowed the Maven/Docker cache key to POM changes, so source-only edits do not invalidate the dependency cache.
- Enabled Maven dependency caching on the macOS and Windows test jobs.
- Added `pull-requests: read` so the integration job can inspect changed files.

Closes #1188 

Note: The changes on this PR were highlighted and addressed with AI assistance

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,7 +32,6 @@ concurrency:
 
 permissions:
   contents: read
-  pull-requests: read
 
 env:
   DOCKER_VOLUME_PREFIX: ".docker/"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,11 +19,8 @@ name: Test
 
 on:
   push:
-    branches:
-      - '**'
-      - '!dependabot/**'
-    tags:
-      - '**'
+    branches: [main]
+    tags: ['**']
   pull_request:
 
 concurrency:
@@ -32,6 +29,7 @@ concurrency:
 
 permissions:
   contents: read
+  pull-requests: read
 
 env:
   DOCKER_VOLUME_PREFIX: ".docker/"
@@ -66,7 +64,7 @@ jobs:
         uses: actions/cache@v5
         with:
           path: .docker
-          key: maven-${{ matrix.jdk }}-${{ matrix.maven }}-${{ hashFiles('compose.yaml', '**/pom.xml', '**/*.java') }}
+          key: maven-${{ matrix.jdk }}-${{ matrix.maven }}-${{ hashFiles('compose.yaml', '**/pom.xml') }}
           restore-keys: maven-${{ matrix.jdk }}-${{ matrix.maven }}-
       - name: Execute Docker Build
         env:
@@ -94,16 +92,17 @@ jobs:
             jdk: 17
             macos: latest
     steps:
-      - name: Set up Java
-        uses: actions/setup-java@v5
-        with:
-          distribution: 'temurin'
-          java-version: ${{ matrix.jdk }}
       - name: Checkout Arrow
         uses: actions/checkout@v6
         with:
           fetch-depth: 0
           submodules: recursive
+      - name: Set up Java
+        uses: actions/setup-java@v5
+        with:
+          distribution: 'temurin'
+          java-version: ${{ matrix.jdk }}
+          cache: 'maven'
       - name: Build
         shell: bash
         env:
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -19,11 +19,8 @@ name: Test
 
 on:
   push:
-    branches:
-      - '**'
-      - '!dependabot/**'
-    tags:
-      - '**'
+    branches: [main]
+    tags: ['**']
```
kou: Could you revert this? If we use this, we can't run CI on fork. All contributors need to open a PR to run CI. It'll increase CI time on apache/arrow-java.

xborder: Reverted. I noticed workflows were duplicated because I created a PR against my own fork. I guess this was a issue just for me

.github/workflows/test.yml
```diff
@@ -149,59 +152,111 @@ jobs:
...
+              filename === 'pom.xml' ||
+              filename === '.github/workflows/test.yml' ||
+              filename.startsWith('c/') ||
+              filename.startsWith('flight/') ||
+              filename.startsWith('format/') ||
+              filename.startsWith('vector/') ||
+              filename.startsWith('ci/scripts/') ||
+              filename.startsWith('testing/data/') ||
+              filename.endsWith('/pom.xml')
+            );
```
kou: Let's separate this job to another workflow and use `paths:` https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onpushpull_requestpull_request_targetpathspaths-ignore .

If you work on this as a separated PR, we'll be able to merge this PR sooner.

.github/workflows/test.yml
```diff
@@ -66,7 +67,7 @@ jobs:
         uses: actions/cache@v5
         with:
           path: .docker
-          key: maven-${{ matrix.jdk }}-${{ matrix.maven }}-${{ hashFiles('compose.yaml', '**/pom.xml', '**/*.java') }}
+          key: maven-${{ matrix.jdk }}-${{ matrix.maven }}-${{ hashFiles('compose.yaml', '**/pom.xml') }}
```
kou, line 69: Why is this needed?
If we don't update a cache by `.java` change, I think that we can't reuse cache for changed `.java`.

xborder: This is used for [maven dependencies](https://github.com/xborder/arrow-java/blob/cb24576e895c0bf8e1d062e2fc82f2b53bfa2eb8/compose.yaml#L47), not for java binaries. It shouldn't change unless you update the dependencies. 
Let me know if I misunderstood something

.github/workflows/test.yml
xborder: I'll revert this and address it in a separate PR then

.github/workflows/test.yml
```diff
@@ -32,6 +32,7 @@ concurrency:
 
 permissions:
   contents: read
+  pull-requests: read
```
kou: Can we revert this now?

kou (approved):
+1

jbonofre (approved):

jbonofre:
I will merge as soon as CI is happy.

xborder:
@jbonofre can you trigger CI again? I've added `-Denforcer.skip=true` to the test script since it seemed to cause `ConcurrentModificationException`. Enforcer is already runs during `build.sh` so I think we can skip that for `test.sh`

jbonofre:
@xborder done. I'm doing a new pass there.

PR #1192, opened 2026-06-18
GH-1190: Reserve view slots for empty view vector values

### LINKED ISSUE #1190: [Java] BaseVariableWidthViewVector.handleSafe under-allocates the view buffer for empty (zero-length) values → IndexOutOfBoundsException
### Describe the bug

`BaseVariableWidthViewVector.handleSafe(int index, int dataLength)` sizes the **view buffer** using the value's *content* length instead of the fixed per-element view size, so an **empty (zero-length) value** written at a full view-buffer boundary is not reallocated and the subsequent `setBytes` reads/writes one element past the end.

```java
// BaseVariableWidthViewVector#handleSafe
protected final void handleSafe(int index, int dataLength) {
    final long targetCapacity = roundUpToMultipleOf16((long) index * ELEMENT_SIZE + dataLength);
    if (viewBuffer.capacity() < targetCapacity) {
      reallocViewBuffer(targetCapacity);
    }
    ...
}
```

Writing view slot `index` requires `(index + 1) * ELEMENT_SIZE` bytes in the view buffer. But the target is `index * ELEMENT_SIZE + dataLength`. For an empty value (`dataLength == 0`) this rounds to `index * ELEMENT_SIZE`, **omitting the slot's own 16 bytes**. When the view buffer is exactly full (`capacity() == index * ELEMENT_SIZE`, e.g. 65536 at `index == 4096`), the `capacity() < targetCapacity` check is false, no reallocation happens, and `setBytes` then executes `viewBuffer.getLong(index * ELEMENT_SIZE)` past the end.

Short non-empty values (`1..INLINE_SIZE`) round up to `index*ELEMENT_SIZE + ELEMENT_SIZE` and reallocate correctly, so the defect is specific to **zero-length** values (e.g. empty strings, and `ViewVarCharWriter`-style null-slot writes that use `EMPTY_BYTES`).

### Reproduction

```java
try (BufferAllocator allocator = new RootAllocator();
     ViewVarCharVector v = new ViewVarCharVector("s", allocator)) {
    v.allocateNew(); // initial 4096-slot / 65536-byte view buffer
    for (int i = 0; i <= 4096; i++) {
        v.setSafe(i, new byte[0]); // empty value
    }
}
```

Throws on `i == 4096`:

```
java.lang.IndexOutOfBoundsException: index: 65536, length: 8 (expected: range(0, 65536))
    at org.apache.arrow.memory.ArrowBuf.checkIndexD(ArrowBuf.java:299)
    at org.apache.arrow.memory.ArrowBuf.getLong(ArrowBuf.java:312)
    at org.apache.arrow.vector.BaseVariableWidthViewVector.setBytes(BaseVariableWidthViewVector.java:1379)
    at org.apache.arrow.vector.BaseVariableWidthViewVector.setSafe(BaseVariableWidthViewVector.java:1183)
```

### Component(s)

Java

### Affected versions

`handleSafe` is byte-identical (and affected) in **18.3.0**, **19.0.0**, and **main** — not a regression of a fixed release.

### Expected behavior

The view buffer should always reserve a full `ELEMENT_SIZE` per slot regardless of content length, i.e. ensure capacity for at least `(index + 1) * ELEMENT_SIZE`. The out-of-line data buffer is sized separately in `setBytes` (`allocateOrGetLastDataBuffer`), so `dataLength` is not needed for the *view* buffer capacity check.

For reference, the C++ and Rust implementations always reserve one fixed-size view per element regardless of content length:

- **C++** `BinaryViewBuilder`: `Reserve(1)` then `data_builder_.UnsafeAppend(BinaryViewType::c_type{})` (and `AppendNull`/`AppendEmptyValue` likewise append a full zeroed 16-byte view).
- **Rust** `GenericByteViewBuilder`: `views_buffer.push(view)` onto a growable `Vec<u128>` (one 16-byte view per value).

A minimal fix would size the view-buffer target to `roundUpToMultipleOf16((long) (index + 1) * ELEMENT_SIZE)` (independent of `dataLength`).

goutamadwant:
## What's Changed

Fixes `BaseVariableWidthViewVector.handleSafe` so `setSafe` reserves a full 16-byte view slot for the target index even when the value length is zero.

Adds coverage for empty values written through the first view-buffer boundary for both `ViewVarCharVector` and `ViewVarBinaryVector`.

## Tests

- `mvn -pl vector -am -Dtest=TestVariableWidthViewVector#testSetSafeEmptyValueAtViewBufferBoundary -Dsurefire.failIfNoSpecifiedTests=false test`
- `mvn -pl vector -am -Dtest=TestVariableWidthViewVector -Dsurefire.failIfNoSpecifiedTests=false test`

Closes #1190.

### DIFF
--- a/vector/src/main/java/org/apache/arrow/vector/BaseVariableWidthViewVector.java
+++ b/vector/src/main/java/org/apache/arrow/vector/BaseVariableWidthViewVector.java
@@ -1405,7 +1405,8 @@ public abstract class BaseVariableWidthViewVector extends BaseValueVector
   }
 
   protected final void handleSafe(int index, int dataLength) {
-    final long targetCapacity = roundUpToMultipleOf16((long) index * ELEMENT_SIZE + dataLength);
+    // The view buffer stores one fixed-width view per value; payload bytes are allocated separately.
+    final long targetCapacity = roundUpToMultipleOf16(((long) index + 1) * ELEMENT_SIZE);
     if (viewBuffer.capacity() < targetCapacity) {
       reallocViewBuffer(targetCapacity);
     }
--- a/vector/src/main/java/org/apache/arrow/vector/BaseVariableWidthViewVector.java
+++ b/vector/src/main/java/org/apache/arrow/vector/BaseVariableWidthViewVector.java
@@ -1405,7 +1405,8 @@ public abstract class BaseVariableWidthViewVector extends BaseValueVector
   }
 
   protected final void handleSafe(int index, int dataLength) {
-    // The view buffer stores one fixed-width view per value; payload bytes are allocated separately.
+    // The view buffer stores one fixed-width view per value; payload bytes are allocated
+    // separately.
     final long targetCapacity = roundUpToMultipleOf16(((long) index + 1) * ELEMENT_SIZE);
     if (viewBuffer.capacity() < targetCapacity) {
       reallocViewBuffer(targetCapacity);
--- a/vector/src/test/java/org/apache/arrow/vector/TestVariableWidthViewVector.java
+++ b/vector/src/test/java/org/apache/arrow/vector/TestVariableWidthViewVector.java
@@ -540,6 +540,26 @@ public class TestVariableWidthViewVector {
     }
   }
 
+  @ParameterizedTest
+  @MethodSource({"vectorCreatorProvider"})
+  public void testSetSafeEmptyValueAtViewBufferBoundary(
+      Function<BufferAllocator, BaseVariableWidthViewVector> vectorCreator) {
+    try (final BaseVariableWidthViewVector vector = vectorCreator.apply(allocator)) {
+      final byte[] emptyValue = new byte[0];
+      vector.allocateNew();
+      final int valueCapacity = vector.getValueCapacity();
+
+      for (int i = 0; i <= valueCapacity; i++) {
+        vector.setSafe(i, emptyValue);
+      }
+
+      vector.setValueCount(valueCapacity + 1);
+      assertTrue(vector.getValueCapacity() > valueCapacity);
+      assertEquals(0, vector.getValueLength(valueCapacity));
+      assertArrayEquals(emptyValue, vector.get(valueCapacity));
+    }
+  }
+
   @Test
   public void testNullableVarType1() {
 
--- a/vector/src/test/java/org/apache/arrow/vector/TestVariableWidthViewVector.java
+++ b/vector/src/test/java/org/apache/arrow/vector/TestVariableWidthViewVector.java
@@ -560,6 +560,40 @@ public class TestVariableWidthViewVector {
     }
   }
 
+  @ParameterizedTest
+  @MethodSource({"vectorCreatorProvider"})
+  public void testSetValueCountFillsEmptiesAtViewBufferBoundary(
+      Function<BufferAllocator, BaseVariableWidthViewVector> vectorCreator) {
+    try (final BaseVariableWidthViewVector vector = vectorCreator.apply(allocator)) {
+      vector.allocateNew();
+      final int valueCapacity = vector.getValueCapacity();
+
+      vector.setSafe(valueCapacity - 1, "x".getBytes(StandardCharsets.UTF_8));
+      vector.setValueCount(valueCapacity + 1);
+
+      assertTrue(vector.getValueCapacity() > valueCapacity);
+      assertTrue(vector.isNull(valueCapacity));
+    }
+  }
+
+  @ParameterizedTest
+  @MethodSource({"vectorCreatorProvider"})
+  public void testSetNullAtViewBufferBoundary(
+      Function<BufferAllocator, BaseVariableWidthViewVector> vectorCreator) {
+    try (final BaseVariableWidthViewVector vector = vectorCreator.apply(allocator)) {
+      vector.allocateNew();
+      final int valueCapacity = vector.getValueCapacity();
+
+      for (int i = 0; i <= valueCapacity; i++) {
+        vector.setNull(i);
+      }
+      vector.setValueCount(valueCapacity + 1);
+
+      assertTrue(vector.getValueCapacity() > valueCapacity);
+      assertTrue(vector.isNull(valueCapacity));
+    }
+  }
+
   @Test
   public void testNullableVarType1() {
 


vector/src/main/java/org/apache/arrow/vector/BaseVariableWidthViewVector.java
```diff
@@ -1405,7 +1405,8 @@ public final int getTotalValueLengthUpToIndex(int index) {
   }
 
   protected final void handleSafe(int index, int dataLength) {
-    final long targetCapacity = roundUpToMultipleOf16((long) index * ELEMENT_SIZE + dataLength);
+    // The view buffer stores one fixed-width view per value; payload bytes are allocated separately.
+    final long targetCapacity = roundUpToMultipleOf16(((long) index + 1) * ELEMENT_SIZE);
```
mbutrovich, line 1410: Correct. Reserves the full slot regardless of payload; `((long) index + 1)` avoids overflow. `roundUpToMultipleOf16` is a no-op here but fine to keep.

vector/src/test/java/org/apache/arrow/vector/TestVariableWidthViewVector.java
```diff
@@ -540,6 +540,26 @@ public void testSizeOfViewBufferElements() {
     }
   }
 
+  @ParameterizedTest
+  @MethodSource({"vectorCreatorProvider"})
+  public void testSetSafeEmptyValueAtViewBufferBoundary(
```
mbutrovich, line 545: Only covers the empty-value path. `setNull` and `fillEmpties`/`setValueCount` reach the same `handleSafe(index, 0)`. Suggest adding tests for those.

### setValueCount trailing gap (silent trigger, no empty string written)
```java
@ParameterizedTest
@MethodSource({"vectorCreatorProvider"})
public void testSetValueCountFillsEmptiesAtViewBufferBoundary(
    Function<BufferAllocator, BaseVariableWidthViewVector> vectorCreator) {
  try (final BaseVariableWidthViewVector vector = vectorCreator.apply(allocator)) {
    vector.allocateNew();
    final int valueCapacity = vector.getValueCapacity();
    vector.setSafe(valueCapacity - 1, "x".getBytes(StandardCharsets.UTF_8));
    // Leaves index valueCapacity unset: fillEmpties -> handleSafe(_, 0) on a full buffer.
    vector.setValueCount(valueCapacity + 1);
    assertTrue(vector.getValueCapacity() > valueCapacity);
    assertTrue(vector.isNull(valueCapacity));
  }
}
```

### setNull at the boundary
```java
@ParameterizedTest
@MethodSource({"vectorCreatorProvider"})
public void testSetNullAtViewBufferBoundary(
    Function<BufferAllocator, BaseVariableWidthViewVector> vectorCreator) {
  try (final BaseVariableWidthViewVector vector = vectorCreator.apply(allocator)) {
    vector.allocateNew();
    final int valueCapacity = vector.getValueCapacity();
    for (int i = 0; i <= valueCapacity; i++) {
      vector.setNull(i);
    }
    vector.setValueCount(valueCapacity + 1);
    assertTrue(vector.getValueCapacity() > valueCapacity);
    assertTrue(vector.isNull(valueCapacity));
  }
}
```

mbutrovich (commented):
Thanks @goutamadwant, this looks like what I was going to do! One coverage gap worth closing: `handleSafe(index, 0)` is also hit by `setNull` (1195) and `fillEmpties` via `setValueCount` (1022), so the bug was reachable without ever writing an empty string. The fix covers those too, but the test only exercises `setSafe(i, new byte[0])`. Suggest adding the two tests below. Non-blocking.

vector/src/main/java/org/apache/arrow/vector/BaseVariableWidthViewVector.java
goutamadwant: Thanks for reviewing @mbutrovich . I kept this code path unchanged since it reserves the full view slot regardless of payload size and the cast keeps the sizing expression safe before the multiplication. Let me know if there any other thoughts.

vector/src/test/java/org/apache/arrow/vector/TestVariableWidthViewVector.java
goutamadwant: @mbutrovich added coverage for both additional paths: `setNull` and `setValueCount`/`fillEmpties` at the view-buffer boundary. The production fix is unchanged and his just closes the missing regression coverage. Let me know. thanks!

goutamadwant:
@mbutrovich addressed your comments. let me know if you have more suggestions. thanks!

mbutrovich (approved):
This LGTM, thanks for the PR and revision @goutamadwant!

parthchandra (approved):
Thanks @goutamadwant. The original implementation is logically wrong and this is the correct fix. 
Also, thanks for adding the tests !

parthchandra:
@goutamadwant could you rebase on main and also run spotless?

parthchandra:
@lidavidm @wgtmac @laurentgo @jbonofre Can you please take a look?

lidavidm (commented):
Seems reasonable, but CI is not passing

parthchandra:
> but CI is not passing

Looks like a rebase and running spotless will fix the CI issues.

goutamadwant:
@parthchandra Rebased this PR on latest main and pushed the update... also ran Spotless; the only formatting change was wrapping the long comment in BaseVariableWidthViewVector.java.

checked locally with below runs : 

- mvn --batch-mode -pl vector spotless:check
- mvn --batch-mode -pl vector -am -Dtest=TestVariableWidthViewVector -Dsurefire.failIfNoSpecifiedTests=false test

The focused vector test passed under both netty and unsafe runs. Let me know. thanks!

goutamadwant:
@parthchandra  any further change needed from my side for this PR ?

parthchandra:
@lidavidm @wgtmac @laurentgo @jbonofre  is anyone able to look into the ci failures? They are unrelated to this PR as far as I can see.

lidavidm:
Please rebase

goutamadwant:
> Please rebase

@lidavidm rebased onto the latest main and pushed the updated commits. thanks!

lidavidm (approved):

PR #1195, opened 2026-06-26
MINOR: Change access modifier of StructVector reallocValidityBuffer

lriggs:
## What's Changed
Changes the access modifier of reallocValidityBuffer from package-private to protected.

This allows subclasses to reallocate the validity buffer independently of the struct elements. Previously, expanding the validity buffer required a full vector reallocation, which could cause an OversizedAllocationException if the child data vectors were already near their capacity.

### DIFF
--- a/vector/src/main/java/org/apache/arrow/vector/complex/StructVector.java
+++ b/vector/src/main/java/org/apache/arrow/vector/complex/StructVector.java
@@ -529,7 +529,7 @@ public class StructVector extends NonNullableStructVector
     super.reAlloc();
   }
 
-  private void reallocValidityBuffer() {
+  protected void reallocValidityBuffer() {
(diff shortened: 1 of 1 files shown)


selvaganesang (approved):

lidavidm:
What's the use case for subclassing StructVector?

lriggs:
> What's the use case for subclassing StructVector?

We have a helper class which wraps StructVector to provide more fine grained control. Maybe @selvaganesang can provide some more details.

selvaganesang:
Provides a mechanism for independent reallocation of embedded vectors to prevent over-expansion of variable-width vectors when the parent struct vector is reallocated. This only mimics the similar concept from ListVector.

lidavidm:
I suppose, why can't the improvements themselves be upstreamed?

PR #1202, opened 2026-07-02
MINOR: Restrict trigger push branch for GitHub Workflow

apupier:
Feature branches rarely need their own CI runs: the code is already tested when a pull request is opened against a release branch. If the push trigger has no branch restriction and pull_request is also configured, every push to a branch with an open PR runs the workflow twice: once for the push and once for the PR synchronisation.

Always give the push trigger an explicit list of branches: this stops branches created from a release branch from inheriting its workflow runs.

see https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=430408443#GitHubActionsRecommendedPractices-Restrictthepushtriggertospecificbranches

## What's Changed

Closes #NNN.

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -19,7 +19,9 @@ name: Dev
 
 on:
   pull_request: {}
-  push: {}
+  push:
+    branches:
+      - main
 
 concurrency:
   group: ${{ github.repository }}-${{ github.ref }}-${{ github.workflow }}


kou:
I think that we don't need this because we always use our fork repositories. In general, we don't create any feature branch in apache/arrow-java.

If we have this, all contributors need to open a PR to run the CI job before it's ready. It will increase CI time in apache/arrow-java.

jbonofre:
I don't think this change is actually required. We don't use feature branch, or similar (always fork).

apupier:
dependabot is creating the branch directly on the repo
image

jbonofre:
@apupier so what?

apupier:
The workflow is currently launched twice for the same commit: one for the push one for the pull request. Only the pull request one is looked at.

The github runners are shared among the whole Apache organization. These days, the limit is regularly reached, causing all Apache projects using github runners waiting in queue. it will allow toreduce the load on the shared pool of github runners for Apache organization

apupier:
> If we have this, all contributors need to open a PR to run the CI job before it's ready. It will increase CI time in apache/arrow-java.

If contributors are using their own fork to launch the CI  job currently, another alternative is to configure to ignore the dependabot branches

kou:
It's acceptable.

We're using it in apache/arrow: https://github.com/apache/arrow/blob/0f6dd60cffce04f448c45314a790788ec624d1a8/.github/workflows/cpp.yml#L21-L24

PR #1204, opened 2026-07-03
MINOR: Restrict trigger push branch for GitHub Workflow

apupier:
Feature branches rarely need their own CI runs: the code is already tested when a pull request is opened against a release branch. If the push trigger has no branch restriction and pull_request is also configured, every push to a branch with an open PR runs the workflow twice: once for the push and once for the PR synchronisation.

Always give the push trigger an explicit list of branches: this stops branches created from a release branch from inheriting its workflow runs.

see https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=430408443#GitHubActionsRecommendedPractices-Restrictthepushtriggertospecificbranches

## What's Changed

Closes #NNN.

Note that I needed to recreate a PR as the previous one was closed https://github.com/apache/arrow-java/pull/1202#issuecomment-4864548916

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -19,7 +19,9 @@ name: Dev
 
 on:
   pull_request: {}
-  push: {}
+  push:
+    branches-ignore:
+      - dependabot/**
 
 concurrency:
   group: ${{ github.repository }}-${{ github.ref }}-${{ github.workflow }}


jbonofre:
Is it not the same as the previous one?

jbonofre:
FYI the previous one has been closed intentionally as not required. Can you elaborate why we would need this one?

apupier:
> Is it not the same as the previous one?

no, this one is ignoring only the dependabot branches as agreed here https://github.com/apache/arrow-java/pull/1202#issuecomment-4864548916 with @kou

apupier:
> FYI the previous one has been closed intentionally as not required. Can you elaborate why we would need this one?

same reason as mentioned here https://github.com/apache/arrow-java/pull/1202#issuecomment-4864005221 after it was closed

kou (approved):
+1

Could you update the PR description based of our PR template?

jbonofre:
@apupier ok, thanks for the details. I was confused 😄

jbonofre (approved):

PR #1218, opened 2026-07-06
GH-1217: validate view data offsets in BaseVariableWidthViewVector

### LINKED ISSUE #1217: Out-of-bounds read for corrupt view offsets in BaseVariableWidthViewVector
### Describe the bug

`ViewVarCharVector`/`ViewVarBinaryVector` store values longer than `INLINE_SIZE` (12 bytes) out of line, encoding a data-buffer index and an offset inline in the view buffer. When a vector is loaded from an IPC stream these fields come straight from the input.

`BaseVariableWidthViewVector` dereferences them verbatim in `getData`, `getDataPointer`, `hashCode`, `copyFromNotNull` and `splitAndTransferViewBufferAndDataBuffer`, e.g.

```java
dataBuffers.get(bufferIndex).getBytes(dataOffset, result, 0, dataLength);
```

Nothing checks that `bufferIndex` is in range or that `dataOffset + dataLength` fits inside the referenced data buffer. A crafted view whose offset/length points past the data buffer produces an out-of-bounds read: with the default bounds checking it throws `IndexOutOfBoundsException`, but with `arrow.enable_unsafe_memory_access=true` (commonly set in production) it reads arbitrary native heap into the returned value.

### Component(s)

Java

Arawoof06:
## What's Changed

Utf8View and BinaryView values longer than 12 bytes live in a separate data buffer, with the data-buffer index and offset stored inline in the view; when a vector is read from an IPC stream those fields come from untrusted input. BaseVariableWidthViewVector dereferenced them without checking that the index is present or that offset+length stays inside the data buffer, so a crafted view reads past the buffer (and returns arbitrary native heap when arrow.enable_unsafe_memory_access is set). Routed every out-of-line dereference (get, getDataPointer, hashCode, copyFrom, splitAndTransfer) through one getValidatedDataBuffer check so a bad view is rejected at the source rather than in each caller.

Closes #1217.

### DIFF
--- a/vector/src/main/java/org/apache/arrow/vector/BaseVariableWidthViewVector.java
+++ b/vector/src/main/java/org/apache/arrow/vector/BaseVariableWidthViewVector.java
@@ -911,7 +911,8 @@ public abstract class BaseVariableWidthViewVector extends BaseValueVector
         final int readBufOffset =
             viewBuffer.getInt(
                 ((long) i * ELEMENT_SIZE) + LENGTH_WIDTH + PREFIX_WIDTH + BUF_INDEX_WIDTH);
-        final ArrowBuf dataBuf = dataBuffers.get(readBufIndex);
+        final ArrowBuf dataBuf =
+            getValidatedDataBuffer(dataBuffers, readBufIndex, readBufOffset, stringLength);
 
         // allocate data buffer
         ArrowBuf currentDataBuf = target.allocateOrGetLastDataBuffer(stringLength);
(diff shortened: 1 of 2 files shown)


Arawoof06:
Could a maintainer add the bug-fix label? The format check is failing on missing labels and I can't set it myself.

lidavidm:
Hi @Arawoof06 sorry for the delay here. The underlying issue was brought to my attention separately and I have an alternate fix at https://github.com/apache/arrow-java/pull/1291.

PR #1219, opened 2026-07-08
GH-477: Fix Flight JDBC DATE(MILLISECOND) parameter binding

### LINKED ISSUE #477: Flight JDBC Driver not setting parameter correctly when parameter type is DATE<MILLISECOND>

Arrow Version 17.0.0

When executing a prepared statement using the Arrow Flight JDBC driver, the driver sets the value for date parameters as DAYS since unix epoch, even when the parameter type returned by the server is DATE(MILLISECONDS).

https://github.com/apache/calcite-avatica/blob/2baa36e493cd90b89a6dec5cb04db0eefe6aef93/core/src/main/java/org/apache/calcite/avatica/util/DateTimeUtils.java#L1344-L1361

The parameter converter below sets the same value regardless of the type. This results in the server receiving an unexpected parameter value.

https://github.com/apache/arrow/blob/6a2e19a852b367c72d7b12da4d104456491ed8b7/java/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/converter/impl/DateAvaticaParameterConverter.java#L33-L43

```
    Properties props = new Properties();
    String connection = "jdbc:arrow-flight-sql://xx:xx/?useEncryption=false";
    String query = "select date from T where date = ?";

    try (Connection con = DriverManager.getConnection(connection, props);
         PreparedStatement ps = con.prepareStatement(query)) {

      ps.setDate(1, Date.valueOf(LocalDate.of(2000, 1, 1)), null);

      try (ResultSet rs = ps.executeQuery()) {
        rs.next();
      }
    }
```

### Component(s)

Java

yujin5701:
## What's Changed

This fixes Flight JDBC prepared statement binding for Arrow `DATE(MILLISECOND)`
parameters.

Avatica represents JDBC `DATE` values as days since the Unix epoch. The Flight
JDBC parameter binder already writes that value correctly for `DateDayVector`,
but it was also writing the same day value directly into `DateMilliVector`,
which expects milliseconds since the Unix epoch.

This now converts epoch days to epoch milliseconds before writing to
`DateMilliVector`.

Tested with `ArrowFlightPreparedStatementTest`.

Closes #477.

### DIFF
--- a/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/converter/impl/DateAvaticaParameterConverter.java
+++ b/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/converter/impl/DateAvaticaParameterConverter.java
@@ -16,6 +16,7 @@
  */
 package org.apache.arrow.driver.jdbc.converter.impl;
 
+import java.util.concurrent.TimeUnit;
 import org.apache.arrow.vector.DateDayVector;
 import org.apache.arrow.vector.DateMilliVector;
 import org.apache.arrow.vector.FieldVector;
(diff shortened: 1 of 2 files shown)


lidavidm:
@yujin5701 can you rebase? It seems CI did not trigger

yujin5701:
Rebased onto the latest main and force-pushed the updated branch.
Could you check whether CI is triggered now?

lidavidm (approved):
LGTM, thanks

PR #1229, opened 2026-07-15
GH-1194: Preserve empty list offset buffers

### LINKED ISSUE #1194: Empty ListVector and LargeListVector can expose offset buffers with writerIndex greater than capacity

`ListVector` and `LargeListVector` can expose an invalid offset buffer state when `valueCount == 0`.

For an empty list vector, the logical offset buffer should still contain the leading offset entry:

- `ListVector`: `(valueCount + 1) * 4 == 4` bytes
- `LargeListVector`: `(valueCount + 1) * 8 == 8` bytes

However, in the empty-vector path, the offset buffer can have:

```text
readerIndex: 0
writerIndex: 4
capacity: 0
```

or the equivalent `writerIndex: 8, capacity: 0` for `LargeListVector`.

This violates the normal buffer invariant:

```text
0 <= readerIndex <= writerIndex <= capacity
```

Downstream consumers that unwrap or serialize the Arrow buffer through Netty can then fail with:

```text
IndexOutOfBoundsException: readerIndex: 0, writerIndex: 4
(expected: 0 <= readerIndex <= writerIndex <= capacity(0))
```

The issue is that `setReaderAndWriterIndex()` sets the offset buffer writer index based on `valueCount * OFFSET_WIDTH`, which is `0` for empty vectors. But list vectors still require one offset slot even when there are no values.

The same issue applies to both:

- `org.apache.arrow.vector.complex.ListVector`
- `org.apache.arrow.vector.complex.LargeListVector`

### Expected behavior

For `valueCount == 0`, the offset buffer should still have enough capacity and readable bytes for the leading zero offset:

```text
(valueCount + 1) * OFFSET_WIDTH
```

So:

- empty `ListVector` should expose at least 4 bytes for offset `[0]`
- empty `LargeListVector` should expose at least 8 bytes for offset `[0]`

The first offset value should be zero.

### Actual behavior

An empty list vector can expose an offset buffer with a non-zero writer index but zero capacity, causing Netty buffer validation to fail when the buffer is unwrapped or consumed.

### Suggested fix

Update `ListVector.setReaderAndWriterIndex()` and `LargeListVector.setReaderAndWriterIndex()` so the offset buffer writer index is based on:

```java
(valueCount + 1) * OFFSET_WIDTH
```

For the `valueCount == 0` case, ensure the offset buffer has enough capacity for the leading zero offset before setting the writer index.

Care should be taken not to shrink the vector's future offset allocation size when allocating this empty sentinel offset buffer.

### Additional context

This was observed downstream in Dremio after upgrading Arrow Java. The failure occurred while sending a record batch containing an empty list vector, where the send path unwraps Arrow buffers through Netty.

The downstream error was:

```text
SYSTEM ERROR: IndexOutOfBoundsException: readerIndex: 0, writerIndex: 4
(expected: 0 <= readerIndex <= writerIndex <= capacity(0))
```

This issue is distinct from #1125. That issue involves `UnionListReader.setPosition` on a post-IPC empty list. This issue is about the offset buffer exported by empty `ListVector` / `LargeListVector` instances having an invalid writer-index/capacity relationship.

prashanthbdremio:
## What's Changed

Preserve the required one-entry offset buffer for empty `ListVector` and `LargeListVector` instances when setting value count and when splitting/transferring zero values. This keeps empty list vectors aligned with the Arrow offset-buffer invariant and avoids exposing buffers whose writer index exceeds capacity.

Adds regression coverage for allocated, unallocated, split-and-transfer, and nested empty list vectors.

Closes #1194.

Tests:
- `mvn -pl vector -am -P=-error-prone -Dmaven.gitcommitid.skip=true -Dsurefire.failIfNoSpecifiedTests=false -Dtest=TestSplitAndTransfer,TestListVector,TestLargeListVector test`
- `mvn -P=-error-prone -Dmaven.gitcommitid.skip=true test`

### DIFF
--- a/vector/src/main/java/org/apache/arrow/vector/complex/ListVector.java
+++ b/vector/src/main/java/org/apache/arrow/vector/complex/ListVector.java
@@ -263,10 +263,12 @@ public class ListVector extends BaseRepeatedValueVector
 
   /** Set the reader and writer indexes for the inner buffers. */
   private void setReaderAndWriterIndex() {
+    final long requiredOffsetBufferCapacity = (long) (valueCount + 1) * OFFSET_WIDTH;
     validityBuffer.readerIndex(0);
     offsetBuffer.readerIndex(0);
     if (valueCount == 0) {
       validityBuffer.writerIndex(0);
+      ensureEmptyOffsetBufferCapacity(requiredOffsetBufferCapacity);
     } else {
       validityBuffer.writerIndex(BitVectorHelper.getValidityBufferSizeFromCount(valueCount));
     }
@@ -274,7 +276,28 @@ public class ListVector extends BaseRepeatedValueVector
     // Both are set to 0 means 0 bytes are written to the IPC stream which will crash IPC readers
     // in other libraries. According to Arrow spec, we should still output the offset buffer which
     // is [0].
-    offsetBuffer.writerIndex((long) (valueCount + 1) * OFFSET_WIDTH);
+    offsetBuffer.writerIndex(requiredOffsetBufferCapacity);
+  }
+
+  private void ensureEmptyOffsetBufferCapacity(long requiredCapacity) {
+    if (offsetBuffer.capacity() >= requiredCapacity) {
+      return;
+    }
+    long previousOffsetAllocationSizeInBytes = offsetAllocationSizeInBytes;
+    ArrowBuf oldOffsetBuffer = offsetBuffer;
+    offsetBuffer = allocateOffsetBuffer(requiredCapacity);
+    final long bytesToCopy = Math.min(oldOffsetBuffer.capacity(), requiredCapacity);
+    offsetBuffer.setBytes(0, oldOffsetBuffer, 0, bytesToCopy);
+
+    final int copiedOffsets = (int) (bytesToCopy / OFFSET_WIDTH);
+    final int requiredOffsets = (int) (requiredCapacity / OFFSET_WIDTH);
+    final int lastCopiedOffset =
+        copiedOffsets == 0 ? 0 : offsetBuffer.getInt((copiedOffsets - 1) * OFFSET_WIDTH);
+    for (int i = copiedOffsets; i < requiredOffsets; i++) {
+      offsetBuffer.setInt(i * OFFSET_WIDTH, lastCopiedOffset);
+    }
+    offsetAllocationSizeInBytes = previousOffsetAllocationSizeInBytes;
+    oldOffsetBuffer.getReferenceManager().release();
   }
 
   /**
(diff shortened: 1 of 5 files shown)


prashanthbdremio:
Could a maintainer please add the `bug-fix` label? This PR fixes GH-1194 and does not require the `breaking-change` label.

prashanthbdremio:
could you please take a look  at it? thanks!
cc: @Yicong-Huang, @lriggs, @akravchukdremio, selva@apache.org

vector/src/main/java/org/apache/arrow/vector/complex/ListVector.java
```diff
@@ -263,18 +263,33 @@ public void exportCDataBuffers(List<ArrowBuf> buffers, ArrowBuf buffersPtr, long
...
+
+  private void ensureEmptyOffsetBufferCapacity(long requiredCapacity) {
+    if (offsetBuffer.capacity() >= requiredCapacity) {
+      return;
+    }
+    long previousOffsetAllocationSizeInBytes = offsetAllocationSizeInBytes;
+    ArrowBuf oldOffsetBuffer = offsetBuffer;
+    offsetBuffer = allocateOffsetBuffer(requiredCapacity);
+    offsetBuffer.setBytes(
+        0, oldOffsetBuffer, 0, Math.min(oldOffsetBuffer.capacity(), requiredCapacity));
```
selvaganesang: If requiredCapacity exceeds oldOffsetBuffer.capacity(), the remaining offset entries cannot be left uninitialized or zeroed — per the Arrow spec the offset buffer must be monotonically non-decreasing, so unfilled entries must be set to the last copied offset value to represent empty lists at those positions.

prashanthbdremio: Addressed in the latest push. The helper now copies existing complete offset entries and fills any newly allocated offset entries with the last copied offset value, for both `ListVector` and `LargeListVector`.

Validation: `mvn -pl vector -am -P=-error-prone -Dmaven.gitcommitid.skip=true -Dsurefire.failIfNoSpecifiedTests=false -Dtest=TestSplitAndTransfer,TestListVector,TestLargeListVector test`

lidavidm:
@prashanthbdremio please rebase, I only just got CI working again

prashanthbdremio:
Rebased onto current `apache/main` and force-pushed the branch.

Validation passed locally:
`mvn -pl vector -am -P=-error-prone -Dmaven.gitcommitid.skip=true -Dsurefire.failIfNoSpecifiedTests=false -Dtest=TestSplitAndTransfer,TestListVector,TestLargeListVector test`

PR #1231, opened 2026-07-15
GH-1230: fix out-of-bounds write in ClobConsumer buffer growth

### LINKED ISSUE #1230: [Java][JDBC] ClobConsumer writes past VarCharVector data buffer for large CLOBs
ClobConsumer copies CLOB data into the VarCharVector data buffer with `MemoryUtil.copyToMemory` at offset `startIndex + totalBytes`, but the buffer-growth guard checks `dataBuffer.writerIndex() + bytes.length > dataBuffer.capacity()`. `copyToMemory` never advances `writerIndex()`, and nothing else does either, so it stays at 0 and the guard only ever checks that a single chunk fits at offset 0. Once the cumulative bytes of a CLOB (or a batch of CLOB rows) exceed the initial data-buffer allocation of about 32 KB, `reallocDataBuffer()` is never called and `copyToMemory`, which performs no ArrowBuf bounds checking, writes past the allocation into adjacent off-heap memory.

The sibling `BinaryConsumer` uses the correct check: `while (vector.getDataBuffer().capacity() < (startOffset + dataLength + read))`.

To reproduce, consume a single CLOB whose UTF-8 length exceeds the initial data buffer (for example `INITIAL_VALUE_ALLOCATION * 8 * 4` characters). The out-of-bounds write corrupts the adjacent offset buffer, and reading the value back throws `NegativeArraySizeException`.

Arawoof06:
## What's Changed

ClobConsumer grows the VarCharVector data buffer with a guard on `dataBuffer.writerIndex() + bytes.length > capacity()`, but it writes each chunk at `startIndex + totalBytes` through a raw `MemoryUtil.copyToMemory` that never advances `writerIndex()`, so the guard stays stuck at offset 0 and `reallocDataBuffer()` is never called for a CLOB whose bytes exceed the initial ~32KB allocation. The unchecked write then runs past the buffer into adjacent off-heap memory. Both consume paths now check `capacity() < startIndex + totalBytes + bytes.length` and re-read the buffer after a realloc, matching the sibling `BinaryConsumer`.

Closes #1230.

### DIFF
--- a/adapter/jdbc/src/main/java/org/apache/arrow/adapter/jdbc/consumer/ClobConsumer.java
+++ b/adapter/jdbc/src/main/java/org/apache/arrow/adapter/jdbc/consumer/ClobConsumer.java
@@ -83,8 +83,9 @@ public abstract class ClobConsumer extends BaseConsumer<VarCharVector> {
             String str = clob.getSubString(read, readSize);
             byte[] bytes = str.getBytes(StandardCharsets.UTF_8);
 
-            while ((dataBuffer.writerIndex() + bytes.length) > dataBuffer.capacity()) {
+            while (dataBuffer.capacity() < startIndex + totalBytes + bytes.length) {
               vector.reallocDataBuffer();
+              dataBuffer = vector.getDataBuffer();
             }
             MemoryUtil.copyToMemory(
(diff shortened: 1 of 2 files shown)


Arawoof06:
gentle ping

lidavidm:
@Arawoof06 please rebase

Arawoof06:
Rebased onto main.

PR #1233, opened 2026-07-17
GH-1232: don't drop TLS for endpoint locations in getStreams

### LINKED ISSUE #1232: [Java][FlightSQL][JDBC] Driver drops TLS for endpoint locations advertised by the server
When a query result has endpoints with non-empty locations, `ArrowFlightSqlClientHandler.getStreams` clones the connection's `Builder` and connects to each advertised location. The clone keeps `username`/`password`, `token` and the OAuth config, and encryption is then set from the location scheme alone:

```java
.withEncryption(endpointUri.getScheme().equals(LocationSchemes.GRPC_TLS))
```

So a location with any other scheme (`grpc+tcp://` in particular) turns encryption off for that endpoint client even when the connection was opened with `useEncryption=true`. `build()` then runs the handshake and sends the credentials over the plaintext channel to the advertised host.

The documented meaning of `useEncryption` (default `true`) is "Whether to use TLS (the default is an encrypted connection)", so a server-supplied string silently overriding it is surprising: a compromised or hostile Flight SQL server, or anything able to influence the `FlightInfo` it returns, can have the driver hand over the user's credentials in cleartext, and a passive attacker on the endpoint path can read them.

Reproduced against a handler built with `withEncryption(true)` plus a username/password, given a `FlightInfo` with one endpoint at `Location.forGrpcInsecure(...)`: the driver attempts the connection and reaches `ClientHandshakeWrapper` on the unencrypted channel instead of refusing it.

`arrow-flight-sql-jdbc-driver`, main.

Arawoof06:
## What's Changed

getStreams clones the connection builder for each advertised endpoint location, and that clone still carries username/password, token and the OAuth config, so whatever it connects to gets the credentials. Encryption for the clone came from the location scheme alone, meaning a `grpc+tcp://` location advertised by the server turned TLS off even for a connection opened with `useEncryption=true`, and the handshake then went out in cleartext to the advertised host. Since the scheme is server-supplied and useEncryption is the user's stated intent, the scheme should not be able to override it downwards; a non-TLS location is now refused through the per-location exception path that is already there for unreachable locations, so the remaining locations are still tried.

Closes #1232.

### DIFF
--- a/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/client/ArrowFlightSqlClientHandler.java
+++ b/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/client/ArrowFlightSqlClientHandler.java
@@ -183,11 +183,22 @@ public final class ArrowFlightSqlClientHandler implements AutoCloseable {
                       sqlClient.getStream(endpoint.getTicket(), getOptions()), null);
               break;
             }
+            final boolean encryptEndpoint =
+                endpointUri.getScheme().equals(LocationSchemes.GRPC_TLS);
+            if (builder.useEncryption && !encryptEndpoint) {
+              exceptions.add(
+                  new SQLException(
+                      String.format(
+                          "Refusing to connect to endpoint location %s without encryption "
+                              + "because the connection to %s is encrypted.",
+                          endpointUri, builder.getLocation().getUri())));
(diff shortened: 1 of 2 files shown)


Arawoof06:
gentle ping

lidavidm:
@Arawoof06 please rebase.

Arawoof06:
Rebased onto main.

PR #1237, opened 2026-07-20
GH-1236: fix out-of-slice read in ArrowFlightJdbcArray.checkBoundaries

### LINKED ISSUE #1236: ArrowFlightJdbcArray.getArray(index, count) can read past the end of the array slice
`ArrowFlightJdbcArray.checkBoundaries` validates the caller-supplied index against `startOffset + valuesCount`:

```java
private void checkBoundaries(long index, int count) {
  if (index < 0 || index + count > this.startOffset + this.valuesCount) {
    throw new ArrayIndexOutOfBoundsException();
  }
}
```

but `index` is relative to the start of the array; both call sites add `startOffset` to it only afterwards, e.g.

```java
checkBoundaries(index, count);
return getArrayNoBoundCheck(
    this.dataVector, LargeMemoryUtil.checkedCastToInt(this.startOffset + index), count);
```

So the accepted range is too large by exactly `startOffset` elements, and `getArray(index, count)` / `getResultSet(index, count)` will read up to that far past the end of the row's slice.

`AbstractArrowFlightJdbcListVectorAccessor` builds these with the offsets of the list element being read, so any row of a list column that does not start at child offset 0 is affected. Reading within the element count the driver itself advertises then returns values belonging to other rows of the shared child vector, and past the child vector's `valueCount` it returns whatever is in allocated-but-unwritten memory.

Reproducer against an `IntVector` of 127 values, with an array covering elements 5..7:

```java
ArrowFlightJdbcArray array = new ArrowFlightJdbcArray(dataVector, 5, 3);
array.getArray(1, 3); // accepted; returns elements 6, 7, 8 — element 8 is outside the array
```

Every existing test constructs the array with `startOffset` 0, where the wrong bound happens to coincide with the correct one, which is why this is not currently caught.

Arawoof06:
## What's Changed

checkBoundaries validates the caller index against startOffset + valuesCount, but index is relative to the start of the array and both call sites add startOffset to it only after the check, so the accepted range is too long by exactly startOffset elements. getArray and getResultSet then read that far past the end of the row's slice. AbstractArrowFlightJdbcListVectorAccessor constructs these from the offsets of the list element being read, so any row of a list column not starting at child offset 0 hands back values belonging to neighbouring rows of the shared child vector, and past the child's valueCount whatever is in allocated-but-unwritten memory. The bound belongs on the relative index, so it is compared against valuesCount instead; that is the same expression as today when startOffset is 0, which is why the existing tests all pass unchanged and only the offset case gets tighter. Added a regression test for each of the two entry points, both of which fail on main.

Closes #1236.

### DIFF
--- a/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightJdbcArray.java
+++ b/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/ArrowFlightJdbcArray.java
@@ -90,7 +90,7 @@ public class ArrowFlightJdbcArray implements Array {
   }
 
   private void checkBoundaries(long index, int count) {
-    if (index < 0 || index + count > this.startOffset + this.valuesCount) {
+    if (index < 0 || index + count > this.valuesCount) {
       throw new ArrayIndexOutOfBoundsException();
     }
   }
--- a/flight/flight-sql-jdbc-core/src/test/java/org/apache/arrow/driver/jdbc/ArrowFlightJdbcArrayTest.java
+++ b/flight/flight-sql-jdbc-core/src/test/java/org/apache/arrow/driver/jdbc/ArrowFlightJdbcArrayTest.java
@@ -106,6 +106,20 @@ public class ArrowFlightJdbcArrayTest {
   }
 
   @Test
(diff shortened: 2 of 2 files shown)


lidavidm:
@Arawoof06 please rebase.

Arawoof06:
Rebased onto main, no conflicts.

PR #1247, opened 2026-07-23
GH-1246: validate decimal byte length before writing in setBigEndian

### LINKED ISSUE #1246: DecimalVector.setBigEndian writes oversized value before validating length
DecimalVector.setBigEndian(int, byte[]) and Decimal256Vector.setBigEndian(int, byte[]) validate the input length only after copying the bytes. On a little-endian platform the method byte-swaps the value into the fixed 16/32-byte slot with a loop of unchecked MemoryUtil.putByte writes:

```java
for (int byteIdx = 0; byteIdx < length; ++byteIdx) {
  MemoryUtil.putByte(outAddress + byteIdx, value[length - 1 - byteIdx]);
}
...
throw new IllegalArgumentException("Invalid decimal value length. Valid length in [1 - 16], got " + length);
```

If value.length exceeds the type width, the loop writes past the slot into adjacent off-heap memory before the IllegalArgumentException is thrown. MemoryUtil.putByte does no bounds checking, so this corrupts memory in the default configuration. setBigEndianSafe(int, long, ArrowBuf, int) performs the same write with no length check at all.

Reproducer: write a value into slot 1, then call setBigEndian(0, new byte[24]) on a DecimalVector; slot 1 is left corrupted (the write spills 8 bytes into it) even though the call throws.

Arawoof06:
## What's Changed

`setBigEndian(int, byte[])` on `DecimalVector` and `Decimal256Vector` byte-swaps the value into the fixed-width slot with unchecked `MemoryUtil` writes and only checks the length afterwards, so a byte array longer than the type width overruns the slot into adjacent off-heap memory before the `IllegalArgumentException` fires. `setBigEndianSafe(int, long, ArrowBuf, int)` does the same write with no length check at all. This moves the length check ahead of the write and adds it to the safe variant, so oversized input is rejected before any memory is touched; valid lengths are unaffected.

Closes #1246.

### DIFF
--- a/vector/src/main/java/org/apache/arrow/vector/Decimal256Vector.java
+++ b/vector/src/main/java/org/apache/arrow/vector/Decimal256Vector.java
@@ -208,6 +208,14 @@ public final class Decimal256Vector extends BaseFixedWidthVector
     BitVectorHelper.setBit(validityBuffer, index);
     final int length = value.length;
 
+    // Reject an oversized value before any bytes are written. The little-endian path below
+    // copies all `length` bytes into the fixed TYPE_WIDTH slot with unchecked native writes,
+    // so validating after the copy would leave adjacent memory already corrupted.
+    if (length > TYPE_WIDTH) {
+      throw new IllegalArgumentException(
+          "Invalid decimal value length. Valid length in [1 - 32], got " + length);
+    }
+
     // do the bound check.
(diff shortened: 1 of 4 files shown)


lidavidm:
@Arawoof06 please rebase.

Arawoof06:
Rebased on main, should be good now.

PR #1248, opened 2026-07-24
GH-1196: Write NO_COMPRESSION_LENGTH sentinel for empty buffers in compress()

### LINKED ISSUE #1196: [Java][IPC] AbstractCompressionCodec.compress() writes prefix=0 for empty buffers, incompatible with C++/Python readers
### Describe the bug

When a buffer has `writerIndex == 0` (e.g., a string column where all values
are empty string ""), `AbstractCompressionCodec.compress()` writes an 8-byte
buffer with `uncompressed_length = 0` as a "shortcut for empty buffer":

https://github.com/apache/arrow-java/blob/main/vector/src/main/java/org/apache/arrow/vector/compression/AbstractCompressionCodec.java#L32-L39

```java
if (uncompressedBuffer.writerIndex() == 0L) {
    // shortcut for empty buffer
    compressedBuffer.setLong(0, 0);  // prefix = 0
    ...
}
```

This has been present since the initial implementation (ARROW-11899, 2021).

The Java `decompress()` handles this correctly (`if size==0 return empty`),
but **C++ and Python Arrow readers do not recognize `prefix=0`**. They attempt
to decompress 0 bytes of data, which fails:

- C++ (Arrow 1.0.0 ~ latest): `IOError: Lz4 compressed input contains less than one frame`
- PyArrow 21.0: same error

### Reproduction

Write an Arrow IPC stream with LZ4_FRAME (or ZSTD) compression where **one
string column has all values = ""** (empty string, not null). The string data
buffer has `writerIndex = 0`, triggering the empty buffer path.

```java
// Writer
ArrowStreamWriter writer = new ArrowStreamWriter(root, null, channel,
    IpcOption.DEFAULT, CommonsCompressionFactory.INSTANCE, CodecType.LZ4_FRAME);

// All rows: stringVector.setSafe(i, "".getBytes());
```

Reading with C++ or Python fails at the first RecordBatch.

### Root cause

The Arrow IPC compression format defines:
- `prefix > 0`: compressed data follows, decompress to `prefix` bytes
- `prefix = -1`: buffer stored uncompressed (sentinel)
- `prefix = 0`: **undefined** — not in spec, not handled by C++/Python

Java writes `prefix=0` for empty buffers, but only Java itself knows how to
read it back. C++/Python treat it as "0 bytes to decompress" → fail.

### Suggested fix

Change the empty buffer path to use `-1` sentinel (which all readers support):

```java
if (uncompressedBuffer.writerIndex() == 0L) {
    ArrowBuf compressedBuffer = allocator.buffer(SIZE_OF_UNCOMPRESSED_LENGTH);
    compressedBuffer.setLong(0, -1L);  // Use -1 instead of 0
    compressedBuffer.writerIndex(SIZE_OF_UNCOMPRESSED_LENGTH);
    uncompressedBuffer.close();
    return compressedBuffer;
}
```

When a reader sees `prefix=-1`, it returns an empty/zero-length slice — which
is correct for an originally empty buffer.

### Environment

- Affected: All Arrow Java versions with IPC compression (1.0.0+)
- Readers that fail: Arrow C++ (all versions), PyArrow (all versions)
- Codec: Both LZ4_FRAME and ZSTD

### Related

- #1116 — similar symptom (prefix=0) but different root cause (race condition
  in vector reuse, not the intentional empty buffer path)
- apache/arrow#15102 — C++ DecompressBuffer fix for prefix=-1 (does not
  handle prefix=0)
```

PG1204:
## What's Changed

`AbstractCompressionCodec.compress()` wrote a length prefix of `0` for empty buffers. The Arrow IPC compression format only defines a positive length (compressed) or `-1` / `NO_COMPRESSION_LENGTH` (stored uncompressed); a `0` prefix is undefined. As a result, C++/PyArrow reject Java-produced streams that contain empty buffers, even though Java itself accepts them on read.

This changes the empty-buffer shortcut in `compress()` to write `CompressionUtil.NO_COMPRESSION_LENGTH` (`-1`) instead of `0`.

`decompress()` is intentionally left unchanged:
- its `NO_COMPRESSION_LENGTH` branch handles the new output, and
- its `decompressedLength == 0` branch remains for backward-compatible reads of older Java-produced streams.

Added a parameterized test (`testEmptyBufferWritesUncompressedSentinel`, covering LZ4 and ZSTD) that compresses an empty buffer, asserts the prefix equals `NO_COMPRESSION_LENGTH`, and asserts it round-trips back to a zero-length buffer. The pre-existing `testEmptyBuffer` continues to pass with empty buffers now routed through the `-1` path.

Closes #1196.

### DIFF
--- a/compression/src/test/java/org/apache/arrow/compression/TestCompressionCodec.java
+++ b/compression/src/test/java/org/apache/arrow/compression/TestCompressionCodec.java
@@ -233,6 +233,35 @@ class TestCompressionCodec {
     AutoCloseables.close(decompressedBuffers);
   }
 
+  static Stream<CompressionCodec> compressingCodecs() {
+    return Stream.of(new Lz4CompressionCodec(), new ZstdCompressionCodec());
+  }
+
+  @ParameterizedTest
+  @MethodSource("compressingCodecs")
+  void testEmptyBufferWritesUncompressedSentinel(CompressionCodec codec) {
+    // GH-1196: an empty buffer must be encoded with the uncompressed-length
+    // sentinel (-1), not 0. A 0 prefix is undefined in the Arrow IPC format and
+    // is rejected by the C++ and Python implementations, even though Java itself
+    // accepts it on read.
+    ArrowBuf empty = allocator.buffer(1);
+    empty.writerIndex(0);
+
+    // compress() takes ownership of and closes the input buffer
+    ArrowBuf compressed = codec.compress(allocator, empty);
+
+    assertEquals(
+        CompressionUtil.NO_COMPRESSION_LENGTH,
+        compressed.getLong(0),
+        "empty buffer must use the -1 uncompressed sentinel, not 0");
+
+    // and it must still round-trip back to an empty buffer
+    ArrowBuf decompressed = codec.decompress(allocator, compressed);
+    assertEquals(0L, decompressed.writerIndex());
+
+    decompressed.close();
+  }
+
   @Test
   void testLz4DecompressRejectsWrongLength() {
     byte[] data = new byte[512]; // all zeros, highly compressible
--- a/vector/src/main/java/org/apache/arrow/vector/compression/AbstractCompressionCodec.java
+++ b/vector/src/main/java/org/apache/arrow/vector/compression/AbstractCompressionCodec.java
@@ -36,7 +36,7 @@ public abstract class AbstractCompressionCodec implements CompressionCodec {
     if (uncompressedLength == 0L) {
       // shortcut for empty buffer
       ArrowBuf compressedBuffer = allocator.buffer(CompressionUtil.SIZE_OF_UNCOMPRESSED_LENGTH);
-      compressedBuffer.setLong(0, 0);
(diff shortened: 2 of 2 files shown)


PG1204:
### Notes for reviewers

- **Empty-buffer path uses `setLong(0, ...)` directly, not the endianness-aware `writeUncompressedLength()`.** This is safe: `reverseBytes(-1) == -1`, so the sentinel reads back correctly on big-endian too. It also matches the existing code, which wrote `0` the same way.
- **Test ownership:** with the `-1` sentinel, `decompress()` now routes through the `NO_COMPRESSION_LENGTH` branch -> `extractUncompressedBuffer()`, which returns a slice sharing the compressed buffer's reference manager. The test closes only the decompressed slice (which releases the shared allocation); closing the original `compressed` buffer as well would be a double-free. Verified leak-free via the `RootAllocator` check in `@AfterEach`.

lidavidm (approved):
Seems reasonable to me.

lidavidm:
@PG1204 can you rebase to pick up CI fixes?

PG1204:
> @PG1204 can you rebase to pick up CI fixes?

@lidavidm rebase done.

PR #1249, opened 2026-07-24
GH-1205: Fix byte-array element leak in FromSchemaByteArray

### LINKED ISSUE #1205: Byte-array elements leak in `FromSchemaByteArray()`

I found a possible JNI array leak in `FromSchemaByteArray()` when the serialized schema cannot be parsed.

`dataset/src/main/cpp/jni_util.cc`

Function: `FromSchemaByteArray`

Relevant code:

```cpp
jbyte* schemaBytes_data =
    env->GetByteArrayElements(schemaBytes, nullptr);
auto serialized_schema = std::make_shared<arrow::Buffer>(
    reinterpret_cast<uint8_t*>(schemaBytes_data),
    schemaBytes_len);
arrow::io::BufferReader buf_reader(serialized_schema);

ARROW_ASSIGN_OR_RAISE(
    std::shared_ptr<arrow::Schema> schema,
    arrow::ipc::ReadSchema(&buf_reader, &in_memo))

env->ReleaseByteArrayElements(
    schemaBytes, schemaBytes_data, JNI_ABORT);
return schema;
```

`GetByteArrayElements()` returns a pointer that must be paired with
`ReleaseByteArrayElements()`.

`ARROW_ASSIGN_OR_RAISE` returns immediately when `ReadSchema()` returns an
error. On that path, the release below the macro is skipped, so the acquired
array elements remain unreleased:

```text
GetByteArrayElements succeeds
  -> ReadSchema returns an error
  -> ARROW_ASSIGN_OR_RAISE returns
  -> ReleaseByteArrayElements is skipped
```

The function is used by the public native `createDataset()` method:

```cpp
schema = JniGetOrThrow(
    FromSchemaByteArray(env, schema_bytes));
```

Malformed, corrupted, or incompatible serialized schema bytes can therefore
reach this path. Repeated failed calls can retain copied array buffers or keep
Java arrays pinned, depending on the JVM implementation.

Suggested fix: release the elements before propagating the parse result, for
example:

```cpp
auto schema_result =
    arrow::ipc::ReadSchema(&buf_reader, &in_memo);

env->ReleaseByteArrayElements(
    schemaBytes, schemaBytes_data, JNI_ABORT);

return schema_result;
```

An RAII guard for `schemaBytes_data` would also ensure release if additional
early returns are introduced later.

PG1204:
## What's Changed

`FromSchemaByteArray()` in `dataset/src/main/cpp/jni_util.cc` acquires the Java byte-array elements via `GetByteArrayElements()`, but the matching `ReleaseByteArrayElements()` was only reached on the success path. When `arrow::ipc::ReadSchema()` failed, `ARROW_ASSIGN_OR_RAISE` returned early and skipped the release, leaking the pinned/copied array buffer on every call with malformed or incompatible serialized schema bytes (e.g. via `createDataset()`).

This releases the elements through an RAII guard so they are always freed on every exit path - success or error, and stays correct if additional early returns are added later.

Closes #1205.

### DIFF
--- a/dataset/src/main/cpp/jni_util.cc
+++ b/dataset/src/main/cpp/jni_util.cc
@@ -352,13 +352,17 @@ arrow::Result<std::shared_ptr<arrow::Schema>> FromSchemaByteArray(
   arrow::ipc::DictionaryMemo in_memo;
   int schemaBytes_len = env->GetArrayLength(schemaBytes);
   jbyte* schemaBytes_data = env->GetByteArrayElements(schemaBytes, nullptr);
+  // Ensure the pinned/copied Java array elements are always released, even when
+  // schema parsing fails and an error Result is returned early below.
+  auto release_elements = [&](jbyte* data) {
+    env->ReleaseByteArrayElements(schemaBytes, data, JNI_ABORT);
+  };
+  std::unique_ptr<jbyte, decltype(release_elements)> elements_guard(schemaBytes_data,
+                                                                    release_elements);
   auto serialized_schema = std::make_shared<arrow::Buffer>(
       reinterpret_cast<uint8_t*>(schemaBytes_data), schemaBytes_len);
   arrow::io::BufferReader buf_reader(serialized_schema);
-  ARROW_ASSIGN_OR_RAISE(std::shared_ptr<arrow::Schema> schema,
-                        arrow::ipc::ReadSchema(&buf_reader, &in_memo))
-  env->ReleaseByteArrayElements(schemaBytes, schemaBytes_data, JNI_ABORT);
-  return schema;
+  return arrow::ipc::ReadSchema(&buf_reader, &in_memo);
 }
 arrow::Status ExportRecordBatch(JNIEnv* env, const std::shared_ptr<RecordBatch>& batch,
                                 jlong struct_array) {
--- a/dataset/src/test/java/org/apache/arrow/dataset/jni/TestFromSchemaByteArray.java
+++ b/dataset/src/test/java/org/apache/arrow/dataset/jni/TestFromSchemaByteArray.java
@@ -28,9 +28,14 @@ import org.junit.jupiter.api.io.TempDir;
 
 /**
  * Regression test for the native {@code FromSchemaByteArray} helper (GH-1205). Passing malformed
- * serialized schema bytes to {@code JniWrapper#createDataset} must surface a Java exception rather
- * than crash, and repeated failures must not leak the pinned/copied Java byte-array elements
- * acquired via {@code GetByteArrayElements}.
+ * serialized schema bytes to {@code JniWrapper#createDataset} exercises the former leak branch in
+ * {@code FromSchemaByteArray} (where {@code ReleaseByteArrayElements} was skipped on error) and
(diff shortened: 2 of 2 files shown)


xborder:
Seems reasonable. From what I was researching, doesn't seem like it is a leak easy to hit and it may only leak a few KBs.

I'd probably just try to add a test that reproduces the issue an ensure that it passes with the fix

PG1204:
> Seems reasonable. From what I was researching, doesn't seem like it is a leak easy to hit and it may only leak a few KBs.
> 
> I'd probably just try to add a test that reproduces the issue an ensure that it passes with the fix

@xborder Agreed it's low-impact and hard to hit. I've added a regression test (TestFromSchemaByteArray) that feeds malformed schema bytes to the native createDataset path, which is the exact branch that used to skip ReleaseByteArrayElements, and confirms it now fails gracefully with the fix in place.

xborder:
@PG1204 I don't think this is testing that we fixed the memory leak. This is just testing the error is thrown. Shouldn't it check no memory was leaked?

PG1204:
> @PG1204 I don't think this is testing that we fixed the memory leak. This is just testing the error is thrown. Shouldn't it check no memory was leaked?

@xborder you're right, that only proves the error path is hit, not that the leak is gone. The catch is what leaks: the buffer wraps the pointer from GetByteArrayElements, which on HotSpot copies the array into a JVM-internal malloc'd buffer, and that copy is what leaked. It's not tracked by NativeMemoryPool, not on the Java heap, and has no portable Java API; it's only visible via process RSS.

So the only way I see to assert "no leak" from Java is an RSS-growth check: loop the call many times with a large payload and assert RSS stays bounded (unfixed balloons by GBs, fixed stays flat). That works on Linux but needs gating on macOS/Windows and is a bit flaky.

Would you prefer (1) that RSS-based test, or (2) dropping the Java assertion and verifying manually with a local leak-checker (valgrind/ASAN), keeping the current test as a graceful-failure guard? Note CI doesn't run a sanitizer today, so option 2 wouldn't be enforced automatically. Happy either way, just want to avoid landing a flaky test.

lidavidm:
I don't think that's reasonable to do. I doubt this is a code path that gets exercised often anyways?

PG1204:
> I don't think that's reasonable to do. I doubt this is a code path that gets exercised often anyways?

agreed, thanks. I'll skip the RSS-based test then, it'd be flaky and isn't worth it for this path. I'll keep the RAII fix plus the existing lightweight test as a graceful-failure regression guard (malformed schema bytes -> exception, no crash), and I'll reword its javadoc so it no longer claims to assert the leak itself. Let me know if you'd rather want to drop the test entirely.

lidavidm:
@PG1204 can you rebase?

(Also, are you using AI assistance at all? I believe ASF may start requiring disclosure soon.)

PG1204:
> @PG1204 can you rebase?
> 
> (Also, are you using AI assistance at all? I believe ASF may start requiring disclosure soon.)

@lidavidm rebase done.
yeah i'm using AI assistance, would you like me to disclose it in the PR description?

PR #1259, opened 2026-08-12
MINOR: actually throw in PooledByteBufAllocatorL

no23reason:
## What's Changed

There was a check in that class that looks like it throws an exception, but actually does not. This means that when that branch is hit, instead of an actionable error, the user gets a ClassCastException like

```
java.lang.ClassCastException: class io.netty.buffer.PooledDirectByteBuf cannot be cast to class io.netty.buffer.PooledUnsafeDirectByteBuf (io.netty.buffer.PooledDirectByteBuf and io.netty.buffer.PooledUnsafeDirectByteBuf are in unnamed module of loader org.springframework.boot.loader.launch.LaunchedClassLoader@5e9f23b4)
```

Actually throwing the exception prevents this and should not break anything: there is still an exception being thrown, just a different type.

Also, make getNormalBufferCount actually return the count, not the size.


lidavidm (approved):
Thanks, can you rebase to pick up CI fixes?

Issue #251, opened 2021-10-26
[Java] Add support for sliced arrays in C Data Interface


asfimport:
The Java implementation of the C Data Interface does not support non-0-offset arrays. This means that arrays like pyarrow.array([0, None, 2, 3, 4]).slice(1, 2) cannot be moved to a Java process. This is not even documented as required by the spec because it was an oversight.

 

**Reporter**: [Roee Shlomo](https://issues.apache.org/jira/browse/ARROW-14474) / @roee88
#### Related issues:
- [[Java] Java implementation of Arrow C data interface](https://github.com/apache/arrow/issues/28685) (is caused by)

<sub>**Note**: *This issue was originally created as [ARROW-14474](https://issues.apache.org/jira/browse/ARROW-14474). Please see the [migration documentation](https://github.com/apache/arrow/issues/14542) for further details.*</sub>

PR #1263, opened 2026-08-19
MINOR: Reject non zero offset on array import

robert3005:
## What's Changed

Instead of silently accepting the array and pretending the import went fine reject arrays where offset is non 0. Add mention to the docs that this is not supported

This is documentation and negative case for #251

### DIFF
--- a/c/src/main/java/org/apache/arrow/c/ArrayImporter.java
+++ b/c/src/main/java/org/apache/arrow/c/ArrayImporter.java
@@ -53,10 +53,6 @@ final class ArrayImporter {
   void importArray(ArrowArray src) {
     ArrowArray.Snapshot snapshot = src.snapshot();
     checkState(snapshot.release != NULL, "Cannot import released ArrowArray");
-    checkState(
-        snapshot.offset == 0,
-        "ArrowArray struct has non-zero offset (%s), which is not supported",
-        snapshot.offset);
 
     // Move imported array
     ArrowArray ownedArray = ArrowArray.allocateNew(allocator);
@@ -87,6 +83,11 @@ final class ArrayImporter {
   }
 
   private void doImport(ArrowArray.Snapshot snapshot) {
+    checkState(
+        snapshot.offset == 0 || snapshot.length == 0,
+        "ArrowArray struct has non-zero offset (%s), which is not supported",
+        snapshot.offset);
+
     // First import children (required for reconstituting parent array data)
     long[] children =
         NativeUtil.toJavaArray(snapshot.children, checkedCastToInt(snapshot.n_children));
--- a/c/src/main/java/org/apache/arrow/c/ArrayImporter.java
+++ b/c/src/main/java/org/apache/arrow/c/ArrayImporter.java
@@ -30,6 +30,7 @@ import org.apache.arrow.vector.FieldVector;
 import org.apache.arrow.vector.dictionary.Dictionary;
 import org.apache.arrow.vector.dictionary.DictionaryProvider;
 import org.apache.arrow.vector.ipc.message.ArrowFieldNode;
+import org.apache.arrow.vector.types.pojo.ArrowType;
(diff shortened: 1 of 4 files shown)


c/src/main/java/org/apache/arrow/c/ArrayImporter.java
```diff
@@ -53,6 +53,10 @@ final class ArrayImporter {
   void importArray(ArrowArray src) {
     ArrowArray.Snapshot snapshot = src.snapshot();
     checkState(snapshot.release != NULL, "Cannot import released ArrowArray");
+    checkState(
+        snapshot.offset == 0,
```
jbonofre: The check throws before the array is moved, so the producer's release callback never fires and the exported data leaks. That's what's turning the six JNI jobs red:

```
[ERROR] RoundtripTest.tearDown:125 » IllegalState Memory was leaked by query. Memory leaked: (48)
```

`ArrowArray.close()` only frees the Java-side struct (it doesn't invoke the release callback) so the buffers `exportCDataBuffers` retained stay retained.

The check also only covers the outermost array, which misses the case that matters most in practice. Suggest dropping it here and re-adding it in `doImport` (see my next comment):

```suggestion
    checkState(snapshot.release != NULL, "Cannot import released ArrowArray");
```

c/src/main/java/org/apache/arrow/c/ArrayImporter.java
jbonofre, line 96: ...and re-add it here instead:

```suggestion
  private void doImport(ArrowArray.Snapshot snapshot) {
    checkState(
        snapshot.offset == 0 || snapshot.length == 0,
        "ArrowArray struct has non-zero offset (%s), which is not supported",
        snapshot.offset);

    // First import children (required for reconstituting parent array data)
```

Three things this position buys us:

1. **The array is released on rejection.** `doImport` runs inside the `try { ... } finally { underlyingAllocation.release(); }` block in `importArray`, so the move has already happened and the producer's release callback fires. That fixes the CI leak, and fixes it for real callers too: `ArrowArrayStreamReader.loadNextBatch` and `NativeScanner.NativeReader.loadNextBatch` both do `try (ArrowArray a = ...) {   Data.importIntoVectorSchemaRoot(...); }`, so as written an offset rejection would strand the producer's entire record batch on the native side.

2. **Children and dictionaries get covered.** `importChild` calls `doImport` as well. This is the case that matters: `ExportRecordBatch` goes through `RecordBatch::ToStructArray()`, which gives the struct offset 0 while each column keeps its own offset. So a sliced pyarrow/Acero `RecordBatch` passes a top-level-only check and still imports from the wrong position: exactly the corruption we're trying to reject. A throw from a child still propagates up to the parent's `finally`, so the release path holds.

3. **`|| snapshot.length == 0`** keeps empty tail slices working. `pa.array([1,2,3]).slice(3, 0)` exports `offset=3, length=0`, and `BufferImportTypeVisitor` sizes every buffer from `fieldNode.getLength()` alone (so those import correctly today, and rejecting them would be a regression on valid input).

jbonofre (changes requested):

c/src/main/java/org/apache/arrow/c/ArrayImporter.java
robert3005: Thanks, I didn't notice that rc/jni and test builds were failing for different reasons

robert3005:
Thanks for the suggestion, I applied it and added test to make sure we don't regress it in the future

jbonofre:
@robert3005 thanks! I'm doing a new pass 😄

jbonofre:
By the way, we merged a fix on CI, I will rebase.

jbonofre:
@robert3005 I updated the `checkState()` to consider `Null` and avoid potential `IllegalStateException`.

I will merge as soon as CI is green.

jbonofre (approved):

lidavidm (approved):

jbonofre (approved):

PR #1282, opened 2026-08-31
GH-1271: [Flight] Reject port 0 in NettyClientBuilder

### LINKED ISSUE #1271: [Flight] Reject port 0 in NettyClientBuilder
Port 0 is reserved and invalid for client TCP connections, but `NettyClientBuilder.build()` currently allows it through (it rejects only negative ports and values above 65535). The lower bound should be tightened to 1.

Four builder-based tests in the JDBC module also have a latent bug where `.withPort()` is never called, so the port field defaults to 0. These fail once the lower bound is tightened:

- `ConnectionTest.testGetBasicClientNoAuthShouldOpenConnection`
- `ConnectionTlsTest.testGetNonAuthenticatedEncryptedClientNoAuth`
- `ConnectionTlsRootCertsTest.testGetNonAuthenticatedEncryptedClientNoAuth`
- `ConnectionMutualTlsTest.testGetNonAuthenticatedEncryptedClientNoAuth`

Each needs `.withPort(FLIGHT_SERVER_TEST_EXTENSION.getPort())` added to its builder call.

This was deferred from #1027 to keep that PR small — see https://github.com/apache/arrow-java/pull/1027#issuecomment-5472093363.

ennuite:
## What's changed

Port validation in `NettyClientBuilder.build()` is tightened to explicitly disallow port 0, as it is not a valid destination port number for a client.
Tests that made use of port 0 were updated to use a valid port under the new validation.

## Are these changes tested?
Yes. `ConnectionTest.testUnencryptedConnectionProvidingInvalidPort` was expanded to cover more invalid ports, including port 0.

Closes #1271.

### DIFF
--- a/flight/flight-core/src/main/java/org/apache/arrow/flight/grpc/NettyClientBuilder.java
+++ b/flight/flight-core/src/main/java/org/apache/arrow/flight/grpc/NettyClientBuilder.java
@@ -140,9 +140,9 @@ public class NettyClientBuilder {
       case LocationSchemes.GRPC_TLS:
         {
           final int port = location.getUri().getPort();
-          if (port < 0 || port > 65535) {
+          if (port < 1 || port > 65535) {
             throw new IllegalArgumentException(
-                "Invalid port " + port + ": must be between 0 and 65535.");
+                "Invalid port " + port + ": must be between 1 and 65535.");
           }
           builder = NettyChannelBuilder.forAddress(location.getUri().getHost(), port);
           break;
--- a/flight/flight-core/src/main/java/org/apache/arrow/flight/grpc/NettyClientBuilder.java
+++ b/flight/flight-core/src/main/java/org/apache/arrow/flight/grpc/NettyClientBuilder.java
@@ -130,7 +130,13 @@ public class NettyClientBuilder {
     return this;
   }
 
-  /** Create the client from this builder. */
+  /**
+   * Create the client from this builder.
+   *
+   * @throws IllegalArgumentException if the URI uses a TCP-based scheme ({@code grpc}, {@code
+   *     grpc+tcp}, {@code grpc+tls}) and the URI contains no port, or a port outside the range [1,
+   *     65535].
+   */
   public NettyChannelBuilder build() {
     final NettyChannelBuilder builder;
 
@@ -140,6 +146,10 @@ public class NettyClientBuilder {
       case LocationSchemes.GRPC_TLS:
         {
           final int port = location.getUri().getPort();
+          if (port == -1) {
+            throw new IllegalArgumentException(
+                "No port specified in location URI: " + location.getUri());
+          }
           if (port < 1 || port > 65535) {
             throw new IllegalArgumentException(
                 "Invalid port " + port + ": must be between 1 and 65535.");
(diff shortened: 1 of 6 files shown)


ennuite:
No need from my side to add this one to release 20.0.0

It needs the enhancement label.

flight/flight-core/src/main/java/org/apache/arrow/flight/grpc/NettyClientBuilder.java
```diff
@@ -140,9 +140,9 @@ public NettyChannelBuilder build() {
       case LocationSchemes.GRPC_TLS:
         {
           final int port = location.getUri().getPort();
-          if (port < 0 || port > 65535) {
+          if (port < 1 || port > 65535) {
```
jbonofre, line 153: `URI.getPort()` returns `-1` to mean "no port present in the URI", not "the user asked for port -1".
Folding it into this range check produces `Invalid port -1: must be between 1 and 65535` (a number that the user never defined).

That path is reachable: a `Location` like `grpc+tls://myhost` with no port, and also `ArrowFlightSqlClientHandler.getStreams()`, which does `withPort(endpointUri.getPort())` on an endpoint URI that may carry no port.

Since this PR is rewriting the message anyway, this seems like the moment to split
the two cases:

```java
if (port == -1) {
  throw new IllegalArgumentException(
      "No port specified in location URI: " + location.getUri());
}
if (port < 1 || port > 65535) {
  throw new IllegalArgumentException(
      "Invalid port " + port + ": must be between 1 and 65535.");
}
```

flight/flight-core/src/main/java/org/apache/arrow/flight/grpc/NettyClientBuilder.java
```diff
@@ -140,9 +140,9 @@ public NettyChannelBuilder build() {
...
             throw new IllegalArgumentException(
-                "Invalid port " + port + ": must be between 0 and 65535.");
+                "Invalid port " + port + ": must be between 1 and 65535.");
```
jbonofre, line 155: This narrows the accepted input range of published public API: `NettyClientBuilder`,
and transitively `FlightClient.Builder`. Third-party code that today calls `FlightClient.builder(allocator, someLocation).build()` with a port-0 location
(a server location captured before `start()`, or a placeholder resolved later) gets a lazily-failing channel today and an immediate `IllegalArgumentException` after this.

The change itself is right, but neither `NettyClientBuilder.build()` nor
`FlightClient.Builder.build()` documents the precondition. 

Could we add an `@throws IllegalArgumentException` javadoc stating the accepted range, so callers can
discover it without hitting it in production?

jbonofre (changes requested):

flight/flight-core/src/main/java/org/apache/arrow/flight/grpc/NettyClientBuilder.java
ennuite: Done!

flight/flight-core/src/main/java/org/apache/arrow/flight/grpc/NettyClientBuilder.java
ennuite: I updated the Javadoc for both methods: `NettyClientBuilder.build()` and `FlightClient.Builder.build()`

Thanks for the review!

jbonofre (approved):
Thanks, LGTM to now!

PR #1289, opened 2026-09-07
GH-551: [Java] Add TimeStamp*TZ support to ComplexCopier

### LINKED ISSUE #551: ComplexCopier is missing cases for `TimeStamp*TZ`

ComplexCopier doesn't support TimeStampNanoTZ or any of the other timezone based types, but looks like BaseWriter might otherwise support the TZ functionality needed.

AzazelSensei:
ComplexCopier skipped types with typeParams except Decimal, so TimeStamp*TZ threw UnsupportedOperationException.

I generate those cases the same way as Decimal, and pass the timezone through to the struct writer. List and struct copy tests for TimeStampNanoTZ.

Closes #551.


AzazelSensei:
Closing this after 15 days without a comment.

PR #1291, opened 2026-09-10
GH-1217: Bounds check before alloc in var-width view vectors

### LINKED ISSUE #1217: Out-of-bounds read for corrupt view offsets in BaseVariableWidthViewVector

lidavidm:
Also, document that when enable_unsafe_memory_access is enabled, all bets are off.

Reported by n0mi1k.

### DIFF
--- a/vector/src/main/java/org/apache/arrow/vector/BaseVariableWidthViewVector.java
+++ b/vector/src/main/java/org/apache/arrow/vector/BaseVariableWidthViewVector.java
@@ -1556,7 +1556,6 @@ public abstract class BaseVariableWidthViewVector extends BaseValueVector
    */
   protected byte[] getData(int index) {
     final int dataLength = getValueLength(index);
-    byte[] result = new byte[dataLength];
     if (dataLength > INLINE_SIZE) {
       // data is in the data buffer
       // get buffer index
@@ -1566,12 +1565,11 @@ public abstract class BaseVariableWidthViewVector extends BaseValueVector
       final int dataOffset =
           viewBuffer.getInt(
               ((long) index * ELEMENT_SIZE) + LENGTH_WIDTH + PREFIX_WIDTH + BUF_INDEX_WIDTH);
-      dataBuffers.get(bufferIndex).getBytes(dataOffset, result, 0, dataLength);
-    } else {
-      // data is in the view buffer
-      viewBuffer.getBytes((long) index * ELEMENT_SIZE + BUF_INDEX_WIDTH, result, 0, dataLength);
+      ArrowBuf dataBuffer = dataBuffers.get(bufferIndex);
+      return dataBuffer.getBytesAsArray(dataOffset, dataLength);
     }
-    return result;
+    // data is in the view buffer
+    return viewBuffer.getBytesAsArray((long) index * ELEMENT_SIZE + BUF_INDEX_WIDTH, dataLength);
   }
 
   protected void getData(int index, ReusableBuffer<?> buffer) {
--- a/vector/src/main/java/org/apache/arrow/vector/BaseVariableWidthViewVector.java
+++ b/vector/src/main/java/org/apache/arrow/vector/BaseVariableWidthViewVector.java
@@ -912,7 +912,8 @@ public abstract class BaseVariableWidthViewVector extends BaseValueVector
             viewBuffer.getInt(
                 ((long) i * ELEMENT_SIZE) + LENGTH_WIDTH + PREFIX_WIDTH + BUF_INDEX_WIDTH);
         final ArrowBuf dataBuf = dataBuffers.get(readBufIndex);
-        if (((long) readBufOffset + (long) stringLength) > dataBuf.capacity()) {
+        if (readBufOffset < 0
+            || ((long) readBufOffset + (long) stringLength) > dataBuf.capacity()) {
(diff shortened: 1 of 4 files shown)


vector/src/main/java/org/apache/arrow/vector/BaseVariableWidthViewVector.java
jbonofre: I'm a bit surprised as #1217 names this method as affected, but it isn't touched by this PR.

`buffer.set(dataBuf, dataOffset, dataLength)` -> `ReusableByteArray.set()` still does `bytes = new byte[len]` before the source `getBytes()` bounds check runs. The same allocate-before-check pattern just fixed two methods above in `getData(int)`.

Should we include the same change here?

jbonofre: Same here: it's mentioned in #1217 but not changed in this PR.

`allocateOrGetLastDataBuffer(viewLength)` allocates a buffer sized by `viewLength` before the bounds-checked `thisDataBuf.setBytes(..., dataBuf, dataOffset, viewLength)` ever runs. A corrupted source view (let's say `copyFrom`/`copyFromSafe` on data loaded from an untrusted IPC stream) drives the same large-allocation DoS this PR fixes in `getData`, through the copy path instead.

jbonofre, line 924: Named in #1217 😄 

`target.allocateOrGetLastDataBuffer(stringLength)` allocated before `currentDataBuf.setBytes(currentOffset, dataBuf, readBufOffset, stringLength)` bounds-checks the source read, and `stringLength` (from `getValueLength(i)`) is never validated first.

This runs on every split/slice/transfer of a View vector, so it's a fairly reachable path for the same bug class this PR addresses elsewhere.

jbonofre, line 1504: Named in #1217 😄 

No array allocation here, but `bufIndex`/`dataOffset` from the view record are used to build an `ArrowBufPointer` with zero validation. Under `arrow.enable_unsafe_memory_access=true`, this is the "reads arbitrary native heap" case, not just a DoS.

jbonofre, line 1524: Same issue as `getDataPointer`: `dataOffset`/`bufIndex` feed `ByteFunctionHelpers.hash(hasher, dataBuf, dataOffset, dataOffset + length)` with no validation.

vector/src/test/java/org/apache/arrow/vector/TestVariableWidthViewVector.java
```diff
@@ -2921,4 +2921,20 @@ public void testValidate() {
       assertTrue(e.getMessage().contains("Not enough capacity for data buffer"));
     }
   }
+
+  @Test
+  public void testValidateInvalidOffsets() {
```
jbonofre, line 2926: `ArrowBuf.setInt(index, value)` takes a byte offset, not a field-slot index. The view record layout is length@0, prefix@4, bufferIndex@8, dataOffset@12 (4 bytes each), but `setInt(1, 0)`, `setInt(2, 0)`, `setInt(3, 1024)` write to byte offsets 1/2/3, overlapping the length field instead of landing on prefix/bufferIndex/dataOffset.

I just tested it: length ends up 64 (unchanged), prefix=4 (collateral overlap), bufferIndex=0, dataOffset=0 (not the intended 1024).

The test passes, but I believe because dataLength=64 exceeds the tiny data buffer's capacity, not because of an out-of-range dataOffset like the test name implies.

I suggest to use `setInt(12, 1024)` to actually cover the corrupted-offset scenario.

vector/src/main/java/org/apache/arrow/vector/BaseVariableWidthViewVector.java
lidavidm: Hmm, adding manual checks everywhere is going to be fairly brittle...

lidavidm: I consolidated all of these methods to use a single accessor that always checks the buffer size.

jbonofre:
@lidavidm thanks for the updates, I'm doing a new pass but it looks good.

PR #1292, opened 2026-09-10
GH-812: Fix appending to an empty UnionVector

### LINKED ISSUE #812: VectorAppender fails with OversizedAllocationException when appending to empty UnionVector

**Description:**
When appending a `UnionVector` to an empty target `UnionVector`, [this](https://github.com/apache/arrow-java/blob/main/vector/src/main/java/org/apache/arrow/vector/util/VectorAppender.java#L556) while loop runs infinitely until the exception occurs:
``` 
org.apache.arrow.vector.util.OversizedAllocationException: Unable to expand the buffer
``` 
It seems to be a issue with the `UnionVector.getValueCapacity()` method. Logging the capacity before and after calling `reAlloc()` shows that it's always `0`. 
Debugging further, it seems like the value capacity of the `internalStruct` in the `UnionVector` always returns `0` because it has no children so [size](https://github.com/apache/arrow-java/blob/main/vector/src/main/java/org/apache/arrow/vector/complex/NonNullableStructVector.java#L373) is `0`.
``` 
 @Override
  public int getValueCapacity() {
    return Math.min(getTypeBufferValueCapacity(), internalStruct.getValueCapacity());
  }
```

**Steps to reproduce** 
```
  @Test
  public void testAppendEmptyTargetUnionVector() {
    final int length = 5;

    try (final UnionVector target = UnionVector.empty("target", allocator);
         final UnionVector delta = UnionVector.empty("delta", allocator)) {

      // populate the delta vector
      delta.setType(0, Types.MinorType.FLOAT4);
      delta.setType(1, Types.MinorType.FLOAT4);
      delta.setType(2, Types.MinorType.FLOAT4);
      delta.setType(3, Types.MinorType.FLOAT4);
      delta.setType(4, Types.MinorType.FLOAT4);

      Float4Vector deltaFloatVector = delta.getFloat4Vector();
      deltaFloatVector.allocateNew();
      ValueVectorDataPopulator.setVector(deltaFloatVector, 1f, 2f, 3f, 4f, 5f);
      assertEquals(length, deltaFloatVector.getValueCount());
      delta.setValueCount(length);

      VectorAppender appender = new VectorAppender(target);
      delta.accept(appender, null);

      assertEquals(length, target.getValueCount());

      for (int i = 0; i < length; i++) {
        Object floatObj = target.getObject(i);
        assertTrue(floatObj instanceof Float);
        assertEquals(i+1, ((Float) floatObj).intValue());
      }
    }
  }
```
This issue does not occur if any value is set in the target vector before appending.

hutiefang76:
## What's Changed

Appending to a newly created `UnionVector` repeatedly reallocates its type buffer until allocation fails: its value capacity remains zero until it has child vectors. Check the type buffer capacity at this stage, since the existing append loop creates and expands the child vectors afterward.

The regression appends two floats to an empty union with a bounded allocator. It fails before the change with an allocation error and passes afterward. All 32 `TestVectorAppender` tests pass with both the Netty and Unsafe allocators on JDK 21. Checkstyle and Spotless also pass.

Closes #812.

Validation (JDK 21):

```sh
mvn -B -pl vector -am test -Dtest=TestVectorAppender -Dsurefire.failIfNoSpecifiedTests=false
mvn -B -pl vector spotless:check
```

Developed with AI assistance; the regression was verified to fail before the fix and pass afterward.

### DIFF
--- a/vector/src/main/java/org/apache/arrow/vector/util/VectorAppender.java
+++ b/vector/src/main/java/org/apache/arrow/vector/util/VectorAppender.java
@@ -601,8 +601,8 @@ public class VectorAppender implements VectorVisitor<ValueVector, Void> {
     UnionVector targetUnionVector = (UnionVector) targetVector;
     int newValueCount = targetVector.getValueCount() + deltaVector.getValueCount();
 
-    // make sure there is enough capacity
-    while (targetUnionVector.getValueCapacity() < newValueCount) {
+    // Child vectors are created and expanded below; an empty union has zero value capacity.
+    while (targetUnionVector.getTypeBuffer().capacity() / UnionVector.TYPE_WIDTH < newValueCount) {
       targetUnionVector.reAlloc();
     }
 
--- a/vector/src/test/java/org/apache/arrow/vector/util/TestVectorAppender.java
+++ b/vector/src/test/java/org/apache/arrow/vector/util/TestVectorAppender.java
@@ -805,6 +805,26 @@ public class TestVectorAppender {
   }
 
   @Test
+  public void testAppendToEmptyUnionVector() {
+    try (BufferAllocator limitedAllocator = allocator.newChildAllocator("empty union", 0, 1048576);
+        UnionVector target = UnionVector.empty("target", limitedAllocator);
+        UnionVector delta = UnionVector.empty("delta", limitedAllocator)) {
+      delta.setType(0, Types.MinorType.FLOAT4);
+      delta.setType(1, Types.MinorType.FLOAT4);
+      Float4Vector values = delta.getFloat4Vector();
+      values.allocateNew();
+      ValueVectorDataPopulator.setVector(values, 1f, 2f);
+      delta.setValueCount(2);
+
+      delta.accept(new VectorAppender(target), null);
+
+      assertEquals(2, target.getValueCount());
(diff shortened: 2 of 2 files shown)


hutiefang76:
The PR-format job failed because the required category label is missing. This is a bug fix for #812 with no breaking API change. Could a maintainer add `bug-fix`? My account does not have permission to add repository labels.

vector/src/main/java/org/apache/arrow/vector/util/VectorAppender.java
```diff
@@ -601,8 +601,8 @@ public ValueVector visit(UnionVector deltaVector, Void value) {
     UnionVector targetUnionVector = (UnionVector) targetVector;
     int newValueCount = targetVector.getValueCount() + deltaVector.getValueCount();
 
-    // make sure there is enough capacity
-    while (targetUnionVector.getValueCapacity() < newValueCount) {
+    // Child vectors are created and expanded below; an empty union has zero value capacity.
```
lidavidm:
```suggestion
```

hutiefang76: Removed in ea411ff. This revision only removes the comment; the implementation and regression test are unchanged.

PR #1294, opened 2026-09-11
GH-1293: Use floor division when splitting epoch millis into day and time

### LINKED ISSUE #1293: [Java][FlightSQL] JDBC driver returns pre-1970 dates one day late when a Calendar is supplied

`DateTimeUtils.getTimestampValue(long)` splits epoch milliseconds into an epoch day and a time within that day. The old code used `/` and `%`:

```java
public static Timestamp getTimestampValue(long millisWithCalendar) {
  long milliseconds = millisWithCalendar;
  if (milliseconds < 0) {
    // LocalTime#ofNanoDay only accepts positive values
    milliseconds -= ((milliseconds / MILLIS_PER_DAY) - 1) * MILLIS_PER_DAY;
  }

  return Timestamp.valueOf(
      LocalDateTime.of(
          LocalDate.ofEpochDay(millisWithCalendar / MILLIS_PER_DAY),
          LocalTime.ofNanoOfDay(TimeUnit.MILLISECONDS.toNanos(milliseconds % MILLIS_PER_DAY))));
}
```

Both operators round toward zero. For dates before 1970, the code adjusted the negative remainder because `LocalTime.ofNanoOfDay` rejects it. It did not adjust the epoch day. When the input was negative and not exactly midnight, the day and remainder then referred to different days. The date came back one day late.

For example, `-618102000000` ms is 1950-06-01 01:00:00 UTC. Division by `86400000` truncates to `-7153`, which is 1950-06-02. The remainder is `3600000` ms, or 01:00. The old result was 1950-06-02 01:00:00. `Math.floorDiv` returns `-7154`, the correct epoch day for 1950-06-01.

### How this happens in JDBC

`ArrowFlightJdbcDateVectorAccessor.getDate(Calendar)` applies the calendar offset before calling this method. A DATE starts at midnight, but any difference between the supplied calendar zone and the JVM default moves it away from midnight. Passing a `Calendar` is the documented way to read a date in a specific zone. As a result, every pre-1970 date read this way came back one day late. Any non-zero offset can expose the bug. Historical offsets can include seconds, such as +05:53:28.

The existing negative test used `-618105600000`, exactly 1950-06-01 00:00:00 UTC. At midnight, truncating division and floor division agree, so that test could not expose the bug.

The fix uses `Math.floorDiv` and `Math.floorMod` for both parts and removes the manual adjustment. Positive values behave as before. A test now covers 1950-06-01 01:00:00 UTC.

This affects `flight-sql-jdbc-core` on `main` and in 19.0.0. It is platform independent. Issues #732 and #324 cover broader timestamp and time-zone behavior. This report is only about the integer-division bug above.

AbhishekPathania:
## What's Changed

`DateTimeUtils.getTimestampValue(long)` used `/` and `%` to split epoch milliseconds into an epoch day and a time within that day. These operators round toward zero. For negative values that were not exactly midnight, the existing code fixed the remainder but not the epoch day. The two parts then referred to different days, so the timestamp came back one day late.

For example, `-618102000000` ms is 1950-06-01 01:00:00 UTC. The old division produced epoch day `-7153`, which is 1950-06-02, while the remainder was 01:00. The method returned 1950-06-02 01:00:00.

This affects DATE values before 1970 when `ArrowFlightJdbcDateVectorAccessor.getDate(Calendar)` applies a non-zero calendar offset. The offset moves the value away from midnight and exposes the division bug.

The method now uses `Math.floorDiv` for the epoch day and `Math.floorMod` for the time within the day. This also removes the need for the manual negative-remainder adjustment. Positive values behave as before.

`DateTimeUtilsTest` now tests 1950-06-01 01:00:00 UTC. The old code returned 1950-06-02 01:00:00 for this input. The test passes with the new floor-based calculation.

Closes #1293.

### DIFF
--- a/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/utils/DateTimeUtils.java
+++ b/flight/flight-sql-jdbc-core/src/main/java/org/apache/arrow/driver/jdbc/utils/DateTimeUtils.java
@@ -55,15 +55,12 @@ public class DateTimeUtils {
    * @return a {@link Timestamp} object representing the given Epoch millis
    */
   public static Timestamp getTimestampValue(long millisWithCalendar) {
-    long milliseconds = millisWithCalendar;
-    if (milliseconds < 0) {
-      // LocalTime#ofNanoDay only accepts positive values
-      milliseconds -= ((milliseconds / MILLIS_PER_DAY) - 1) * MILLIS_PER_DAY;
-    }
-
+    // Millis are negative before 1970, where only floor semantics keep the epoch day
+    // and the time-of-day remainder on the same day (and the remainder non-negative).
     return Timestamp.valueOf(
         LocalDateTime.of(
-            LocalDate.ofEpochDay(millisWithCalendar / MILLIS_PER_DAY),
-            LocalTime.ofNanoOfDay(TimeUnit.MILLISECONDS.toNanos(milliseconds % MILLIS_PER_DAY))));
+            LocalDate.ofEpochDay(Math.floorDiv(millisWithCalendar, MILLIS_PER_DAY)),
+            LocalTime.ofNanoOfDay(
+                TimeUnit.MILLISECONDS.toNanos(Math.floorMod(millisWithCalendar, MILLIS_PER_DAY)))));
   }
 }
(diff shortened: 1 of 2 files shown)


AbhishekPathania:
The "Ensure PR format" check is failing only on the missing label step. I tried to add `bug-fix` when opening the PR, but label changes need write access on the repo (403 for an outside contributor). Could a committer add `bug-fix`? The title and the issue link both pass.

lidavidm (approved):

lidavidm:
Hmm, RC CI is failing, but I think that's not related to this.

PR #1306, opened 2026-09-21
GH-1305: [CI] Drop x86_64 macOS from JNI builds

### LINKED ISSUE #1305: [CI] Drop x86_64 macOS from JNI builds

Homebrew treats x86_64 macOS as a [tier 3 configuration](https://docs.brew.sh/Support-Tiers#tier-3) and rarely builds bottles for it. This has broken the `JNI macos-15-intel x86_64` job in the RC workflow repeatedly, because `brew bundle` doesn't build from source and aborts with `no bottle available!` before anything is compiled:

* #1269 worked around it for `aws-sdk-cpp` and gRPC.
* #1270 worked around it for Node.js.

Each of these was a formula-by-formula patch, and any new dependency added to `apache/arrow`'s `cpp/Brewfile` can break the job again the same way. We can't keep an x86_64 macOS JNI build green on a configuration Homebrew doesn't support.

As noted in #1269, dropping the `macos-15-intel` entry needs its own discussion because it is user-facing, so this issue is that discussion.

### Proposal

Drop x86_64 macOS from CI:

* `rc.yml`: remove the `macos-15-intel` / `x86_64` entry from the `jni-macos` matrix, and stop extracting and verifying `jni-macos-x86_64.tar.gz` in the `binaries` job.
* `test.yml`: remove the `AMD64` / `macos-15-intel` entry from the `macos` test matrix.
* Remove the Homebrew workarounds that only existed for the x86_64 runners (the `/usr/local` `python@XXX --overwrite` conflict loop, and the `uname -m` guard around the `pkg-config` uninstall).

### User-facing impact

Release JARs would no longer bundle x86_64 macOS `.dylib` files for `arrow_cdata_jni`, `arrow_dataset_jni`, `arrow_orc_jni` and `gandiva_jni`. Intel Mac users of those modules would need to build the JNI libraries themselves. `ci/scripts/jni_macos_build.sh` still handles `x86_64`, so a local Intel build keeps working.

aarch64 macOS, x86_64 and aarch64 Linux, and x86_64 Windows are unaffected.

### Component(s)

Java, CI

jbonofre:
### Rationale for this change

Homebrew treats x86_64 macOS as a [tier 3 configuration](https://docs.brew.sh/Support-Tiers#tier-3) and rarely builds bottles for it. `brew bundle` doesn't build from source, so the `JNI macos-15-intel x86_64` job in the RC workflow aborts in `Install dependencies` with `no bottle available!` before anything is compiled.

We've been patching this formula by formula:

* #1269 skipped `aws-sdk-cpp` and gRPC.
* #1270 skipped Node.js.

Any new dependency added to `apache/arrow`'s `cpp/Brewfile` can break the job again the same way, and nightly RC runs go red on `main` when it does. We can't keep this job green on a configuration Homebrew doesn't support.

#1269 noted that dropping the `macos-15-intel` entry needed its own discussion because it's user-facing. #1305 is that discussion.

`rc.yml`:

* Removed the `macos-15-intel` / `x86_64` entry from the `jni-macos` matrix.
* `binaries` no longer extracts `jni-macos-x86_64.tar.gz` nor asserts the four `jni/*/x86_64/*.dylib` files.
* Dropped the `/usr/local` `python@XXX --overwrite` conflict loop. Its comment already said it could go once we dropped `macos-15-intel`, since `macos-14` and later on arm64 use `/opt/homebrew`. `brew update` is kept.
* Unwrapped the `pkg-config` uninstall from its `if [ "$(uname -m)" = "arm64" ]` guard, which is now always true.
* Trimmed the tier-3 justification from the `HOMEBREW_BUNDLE_BREW_SKIP` comment. The skip list itself is unchanged: `aws-sdk-cpp` and gRPC are still skipped so the bundled static ones are used, and Node.js is still skipped because the JNI build doesn't need it.

`test.yml`:

* Removed the `AMD64` / `macos-15-intel` entry from the `macos` test matrix, leaving `AArch64` on `macos-latest`.

`ci/scripts/jni_macos_build.sh` is untouched, so building the JNI libraries locally on an Intel Mac still works.

### Are these changes tested?

Yes, by CI. The remaining `JNI macos-14 aarch_64` job exercises the changed `Install dependencies` step, and `binaries` exercises the changed artifact extraction and verification.

Yes. Release JARs no longer bundle x86_64 macOS `.dylib` files for `arrow_cdata_jni`, `arrow_dataset_jni`, `arrow_orc_jni` and `gandiva_jni`. Intel Mac users of those modules will need to build the JNI libraries themselves. aarch64 macOS, x86_64 and aarch64 Linux, and x86_64 Windows are unaffected.

Closes #1305.


lidavidm (approved):

kou (approved):
+1

